PR #16, opened 2015-07-21
Gremlin php

dmill-bz:
updated PR against develop.


dmill-bz:
I think this can be merged and then refactored as we go at this stage. All tests pass locally and I can help in setting up travis to get the tests running there as well if needed.

Issue #17, opened 2015-07-23
Exceptions


dmill-bz:
Spider should have it's own set of exceptions. (so we can for example convert the Exceptions thrown by the drivers). 
What Exception classes should we add to Spider?

Here are some ideas:
- [ ] `InvalidConfigurationException` - would this be the same as `ItemNotFound` thrown by `Manager`? or for a user who tries to set a `bool` config as a `string`?
- [x] `FormattingException` - If requesting a format not registered. Let's use a registration system to add formats so it can be extended.
- [x] `ServerException` - would this include all server errors? (invalid credentials, no db found, etc?)
- [x] `ConnectionNotFoundException` - already exists. if a non-existing connection is `make()` by the Manager
- [x] `InvalidCommandException` - if the Command or CommandBag is corrupt.
- [x] `NotSupportedException` - For cases where functionalities in some graph DBs aren't supported (transactions?)

Any other ideas? Also what extra functionality do we want to attribute to these (if any)

electricjones:
Tossed in some thoughts. Can't think of any extra functionality.

dmill-bz:
checked some of these as they are in the PR #20

dmill-bz:
For `ServerException` should we get the drivers to catch all the server exceptions thrown by the clients and return this one exception. Or should we let those stay as they may be more explicit (at least during development)

electricjones:
During development, I want to see the client exception. Maybe later we can have them catch and wrap or something.

electricjones:
What do we want to do with this issue? Move it to documentation?

dmill-bz:
Yeah we've done most of it. And what's left we can just implement as we need them. Worst case scenario we make two PRs with the same added Exception but that's an easy conflict to resolve.

Issue #13, opened 2015-07-21
Possible refactoring ideas


dmill-bz:
Thought this might be a good place to put some refactoring ideas we have as things develop:
- [x] **`Object` class** : 
  Currently only Drivers take advantage of dynamic property setting via construct. Maybe we should apply this to all classes? I can definitely see Records using this, possibly even commands. Perhaps we should introduce an `Object` class that will father all other classes and provide this functionality. This would be a good place to place some getter and setter functionality as well. (see #12 )

I have an idea for this at the end of #12

dmill-bz:
can we safely close this?

electricjones:
Yeah, we can always open a new one for small refactors

Issue #11, opened 2015-07-21
Data Mapping to records. Which layer.


dmill-bz:
Here's a recap of previous conversations:

---

The mapping of DB results to objects in the driver just seems out of place. It shouldn't be the driver's place to know what the query is returning. 
In the following case:

``` php
$query = UserRecord::select();
$userRecords = $query->all();
```

It's the Query that gets the `all()`, `one()`, `tree()`, etc. calls. It should be the query's job to populate the records (as it also knows which record spawned it). I think the `SpiderRecord` in the driver is too premature, or at least not properly implemented. 

Maybe the `Driver` should implement multiple formatting methods for different result sets. path, tree, vertices, properties, scalar and the query should call those appropriately.

For instance, `$query->all()` would in reality do:

``` php
$result = $query->execute();
$result = Driver::FormatAll($result);
// Do mapping of results to objects
$return $objectArray;
```

(Doesn't have to be exactly like this but it's for the idea)

---

**Mapping Results**

Mapping results to `SpiderRecord`s in the `Driver` has always felt awkward to me. I've tried it in other places like the `Connection` and on its own.

The problem is twofold. First, people should be able to use these component parts independently. Different layers of abstraction will build on top of each other. That is, use the `Driver` without using the `CommandBuilder`. Once Models come around, they should be able to use the `CommandBuilder` without using the `Driver` directly.

With that in mind, the `Driver` MUST return something consistent. In the orient driver, I am using a third-party language binding and I expect most drivers to do the same. Those bindings will each return their own flavor of records. If it is the `CommandBuilder` or `Connection` or something else that is responsible for mapping, I don't know how to consistently map an unknown record type to a consistent SpiderRecord.

Maybe each driver could/should contain a `RecordTransformer` of some kind? If this one needs more discussion, lets open a new issue and copy some of this.

dmill-bz:
I guess a few questions need answers.

**What are the most common response types?**
- Set of elements
- Single element (should generally be the same as above just with one entry)
- Set of properties (single element)
- Set of set of properties (multiple elements)
- Tree
- Path
- Scalar
- Custom

**Which of these should we handle (ignoring the others)?**
- Set of elements
- Single element (should generally be the same as above just with one entry)
- Tree
- Path
- Scalar

**We need to have the Driver return consistent data across DBs but what type is that?**

Should it be a PHP native formatted in a specific way? I don't think it should be a `Record` because the Driver could be used on it's own without records.
[...]

PR #10, opened 2015-07-20
initial implementation of gremlin-server driver

dmill-bz:
This code is functional and the tests pass (though the tests aren't more rigorous than the Orient ones)

This could be optimized and I made a post in one of the issues with some points I wanted to bring up. This PR is the code to go with those points. 
I might create an issue for each point so we can follow them separately and go into more detail.

PR #20, opened 2015-07-23
Made changes to how the driver data mapping is handled.

dmill-bz:
This is going to need a full review/discussion. I will start writing a complete description of changes to see if we decide to adopt all of them or only a subset.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
     ],
     "require": {
       "php" : ">=5.4.0",
-      "michaels/data-manager": "dev-develop",
+      "michaels/data-manager": "0.8.*",
       "ostico/phporient": "dev-master",
     "brightzone/gremlin-php": "*"
 
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,8 @@
       "php" : ">=5.4.0",
       "michaels/data-manager": "dev-develop",
       "ostico/phporient": "dev-master",
-    "brightzone/gremlin-php": "*"
+    "brightzone/gremlin-php": "*",
+    "everyman/neo4jphp": "dev-master"
 
     },
     "require-dev": {
--- a/src/Base/Collection.php
+++ b/src/Base/Collection.php
@@ -1,0 +1,25 @@
+<?php
+namespace Spider\Base;
+
+use Michaels\Manager\Traits\ManagesItemsTrait;
+use Michaels\Manager\Traits\ChainsNestedItemsTrait;
+
+/**
+ * Class Colection
+ * @package Spider\Base
+ */
+class Collection extends Object
+{
+     /**
+     * @inherits from Michaels\Manager:
+     *      init(), add(), get(), getAll(), exists(), has(), set(),
+     *      remove(), clear(), toJson, isEmpty(), __toString()
+     */
+    use ManagesItemsTrait;
+    use ChainsNestedItemsTrait;
+
+    /**
+     * @var array The item collection
+     */
+    protected $items;
+}
--- a/src/Base/Object.php
+++ b/src/Base/Object.php
@@ -1,0 +1,49 @@
+<?php
+namespace Spider\Base;
+
+/**
+ * Class Object
+ * @package Spider\Base
+ */
+class Object {
+
+    /**
+     * Overriding construct to map configuration to properties
+     *
+     * @param array $configuration the properties to populate
+     *
+     * @return void
+     */
+    public function __construct(array $configuration = [])
+    {
+        $this->setProperties($configuration);
+    }
+
+    /**
+     * Map array to class properties
+     *
+     * @param array $properties the key value array to set
+     *
+     * @return void
+     */
+    public function setProperties(array $properties)
+    {
+        foreach($properties as $key => $value)
+        {
+            $this->setProperty($key, $value);
+        }
+    }
+
+    /**
+     * Set a class property
+     *
+     * @param String|Int $name  the name of the property to set
+     * @param mixed      $value the value of the property
+     *
+     * @return void
+     */
+    public function setProperty($name, $value)
+    {
+        $this->$name = $value;
+    }
+}
--- a/src/Base/Object.php
+++ b/src/Base/Object.php
@@ -12,6 +12,7 @@ class Object {
      *
      * @param array $configuration the properties to populate
      *
+     * @return void
      */
     public function __construct(array $configuration = [])
     {
--- a/src/Base/Object.php
+++ b/src/Base/Object.php
@@ -12,7 +12,6 @@ class Object {
      *
      * @param array $configuration the properties to populate
      *
-     * @return void
      */
     public function __construct(array $configuration = [])
     {
--- a/src/Connections/Connection.php
+++ b/src/Connections/Connection.php
@@ -38,7 +38,7 @@ class Connection implements ConnectionInterface
      */
     public function open()
     {
-        $this->driver->setCredentials($this->getAll()); // from given properties
+        $this->driver->setProperties($this->getAll()); // from given properties
         return $this->driver->open();
     }
 
--- a/src/Drivers/AbstractDriver.php
+++ b/src/Drivers/AbstractDriver.php
@@ -27,6 +27,6 @@ abstract class AbstractDriver extends Collection implements DriverInterface
             $this->StopTransaction(FALSE);
         }
         //close driver
-//        $this->close();
+        $this->close();
     }
 }
--- a/src/Drivers/AbstractDriver.php
+++ b/src/Drivers/AbstractDriver.php
@@ -29,5 +29,4 @@ abstract class AbstractDriver extends Collection implements DriverInterface
         //close driver
         $this->close();
     }
-
 }
--- a/src/Drivers/AbstractDriver.php
+++ b/src/Drivers/AbstractDriver.php
@@ -5,6 +5,18 @@ use Spider\Base\Collection;
 
 abstract class AbstractDriver extends Collection implements DriverInterface
 {
+    /**
+     * set of possible formats for responses.
+     */
+    const FORMAT_SET = 10;
+    const FORMAT_TREE = 20;
+    const FORMAT_PATH = 30;
+    const FORMAT_SCALAR = 40;
+    const FORMAT_CUSTOM = 50;
+
+    /**
+     * @var bool whether or not the driver is currently handling an open transaction
+     */
     public $inTransaction = FALSE;
 
     public function __destruct()
--- a/src/Drivers/AbstractDriver.php
+++ b/src/Drivers/AbstractDriver.php
@@ -27,6 +27,6 @@ abstract class AbstractDriver extends Collection implements DriverInterface
             $this->StopTransaction(FALSE);
         }
         //close driver
-        $this->close();
+//        $this->close();
     }
 }
--- a/src/Drivers/AbstractDriver.php
+++ b/src/Drivers/AbstractDriver.php
@@ -1,58 +1,21 @@
 <?php
 namespace Spider\Drivers;
 
-abstract class AbstractDriver implements DriverInterface
+use Spider\Base\Collection;
+
+abstract class AbstractDriver extends Collection implements DriverInterface
 {
-    /**
-     * Constructs a driver from a properties array
-     * @param array $properties
-     */
-    public function __construct(array $properties = [])
-    {
-        $this->setCredentials($properties);
-    }
+    public $inTransaction = FALSE;
 
-    /**
-     * Sets the credentials from a properties array
-     * @param array $properties
-     */
-    public function setCredentials(array $properties = [])
+    public function __destruct()
     {
-        foreach ($properties as $property => $value) {
-            $this->setCredential($property, $value);
+        //rollback changes
+        if($this->inTransaction)
+        {
+            $this->StopTransaction(FALSE);
         }
+        //close driver
+        $this->close();
     }
 
-    /**
-     * Sets and individual credential configuration item
-     * @param $property
-     * @param $value
-     * @return $this
-     */
-    public function setCredential($property, $value)
-    {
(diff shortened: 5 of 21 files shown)


dmill-bz:
Please bear with me as this is going to be pretty long. It's basically a list of checks and potentially new issues to create.

Check them if you feel they're acceptable without further changes. Once we've gone through them all it should be comfortable to merge:

**Changes :**
- [x] Added a `Spider\Base\Object` class that handles `__construct()` property mapping. It has methods `setProperties()`, and `setProperty()` for the time being.
- [x] Added a `Spider\Base\Collection` class that extends `Object` and implements `data-manager` features.
- [x] Replaced `setCredentials()` where it was used with `setProperties()`
- [x] Removed all credential handling from `Spider\Drivers\AbstractDriver` and had it extend `Collection` instead. Did the same (removal) for `DriverInterface`
- [x] Added Transactions and formatting methods to `DriverInterface` as well as the gremlin driver and added the testing.
- [x] `AbstractDriver` will rollback transactional changes on destruct if the transaction is still open. Forces users to manually commit their transactions. (it will also close the driver as this may free threads for the server in case of sessions)
- [x] When possible I made Driver properties (such as `port` or `hostname` (etc) public as I felt this was more intuitive)
- [x] Created a `Spider\Drivers\Response` class that extends `Collection`. This one is quite straightforward, you can check the source. 
  
  This is the object that every driver should return. OrientDriver will need to be changed.
  You will need to initialize the response like this:
  
  ``` php
  $response = new \Spider\Drivers\Response(['_raw' => $rawResponse, '_driver' => $this]);
  ```
  
  You've got an example in `Spider\Drivers\Gremlin\Driver`. I'm comfortable with this but maybe you would like to do it differently?
- [x] I've added the new transaction and formatting methods to your driver stubs. These methods are all empty though. I figured you would sort it out while refactoring the test suit
- [x] I've readjusted existing tests like `ConnectionTest` and `XDriverTest` to match the new features. Check source here to see if you're ok with it.
- [x] `Driver::FormatToSet()` currently returns the following consistent data (let me know if you're ok with this):
  
  For a single entry:
  
  ``` php
  $consistent = Response::getSet();
  $consistent instanceof \Spider\Base\Collection; // true
  $consistent->meta()->id; // id
  $consistent->meta()->label; // label
  $consistent->meta()->type; // type
  $consistent->someProperty; // the property
  ```
  
  Multiple entries return an array of the above `Collection`s.
- [x] `Response::formatToScalar()` currently returns a single int or string.
- [x] `Response::formatToPath()` currently returns the following consistent data:
  
  ``` php
  [
     [Collection, Collection, ...],
     [Collection, Collection, ...],
     ...
  ]
  ```
  
  Each line in this array is essentially a `Set`
- [x] `Response::formatToTree()` Haven't implemented this yet but we need to agree on the format it should return.

**Todo :**
- [ ] **!!! Fix OrientDriver implementation as I have hardly touched this and tests will fail !!!**
- [x] Have other classes extend `Collection` or `Object`
- [x] Gremlin `Driver` does not check that the responses passed to the `FormatToX()` methods are in the right format. 
- [x] Agree on a format for `Tree`
- [x] No catching of Exceptions. Maybe we can take care of this with #17 instead
- [x] Need to document these changes. Especially `Response` and response formats.

That's about it. I'll let you have a look. let me know what you think/find. Depending on the feedback We can change this around a bit, maybe do a couple of todos and place the rest in their own issues.

electricjones:
This all looks like great work. I have marked the items that I feel are fine as is.

**General**
- Driver should manually ensure that objects passed to `formatToX()` is valid and throw an exception. Discuss on #17 

**A couple thoughts on formatToX**
_General_
- What about `formatAsX()`?
- What should happen if the user tries to format a `Response` to an impossible format for that response. For instance, if the query produced a single Vertex and they tried `formatToPath()`. An exception of some kind or a path of one?
- I don't want to get type crazy, and I like the idea of Drivers _only_ returning `Response`s, but is there a way to type-check a formatted result? Maybe a `Response::getFormat()` that returns a string of the current format? Defaults to something like `"raw"`, `"result"` or even `false`

_Sets_
- Sets look great to me. A set would be a single Vertex or Edge, yes?
- I would probably consider a set to be a Record analogous to a row in SQL,
- Since `id` and probably `label` are so common, would we want to duplicate them at the top-level of the set? `$consistent->id;` as well as `consistent->meta()->id`. This can be done easily enough with magic methods.
- Also, what is the difference between `label` and `type`?
- What is immutable in sets? Do we care about protecting them? I probably don't, but that could lead to problems once we get into Models. If someone changes the `id` and tries to save it, bad things happen. My vote is to let Sets be mutable, but provide id and label protection in the base `Model`

**Documentation**
- [ ] I have initiated a spider/docs repo and am currently setting up with mkdocs and sami for api documentation. I think most heavy documentation can go there and post to readthedocs.com. The readme can be a simple getting started.

**Minor todos after merge**
- phpdocs and style to PSR (my IDE can do that automatically once merged)
- This may have invalidated the `Graphs` directory all together, unless it becomes used for Models. In either case, I think we should drop `Record` so it will not be confused with Sets

**Possible new Issues**
- Should we add helper methods to `Collection` like `pop()`, `sort()`, and such. I have actually considered adding a Collection trait to michaels/manager. See chrismichaels84/data-manager#10 
- Enhance Collections with arrayable functionality. Also a data-manager trait. See chrismichaels84/data-manager#11
- Discuss `Tree` format. Currently Orient doesn't support this natively (that I can find), and it may require two queries for the desired results. Does Gremlin? Or this may be a feature to leave for down the road.

**What I will do**
- [ ] I will update the OrientDriver tonight or tomorrow morning.
- [ ] I have a working branch for extending other classes with `Manager` it should be easy to extend `Collection` or `Object` instead. Will do this after merging this PR.
- [x] I have also pushed a branch completing a very basic CommandBuilder with CRUD operations on single records.

dmill-bz:
**What about `FormatAsX` ?**
Yeah much better! lets go with that.

**What should happen if the user tries to format a Response to an impossible format for that response. For instance, if the query produced a single Vertex and they tried formatToPath(). An exception of some kind or a path of one?**
Ideally it would throw an `Exception`. The problem is that I'm not sure that every driver would be capable of telling the difference between formats. For example a set of vertices and a path might be formatted in the same way. In that situation all bets are off and you need to know what you're doing. 
But whenever possible the driver should throw an `IncorrectFormatException`.

**is there a way to type-check a formatted result?**
This kinda falls into the above comment. Basically we can't guaranty that the driver will know. So we can either put that in with a potential for "I don't know" type returns in some drivers. Or we decide not to.
Maybe we can leave this to later and see if it would help us to have this while mapping the data.

**Sets look great to me. A set would be a single Vertex or Edge, yes?**
A Set would be either a single element (V or E) or multiple elements. Essentially it's the response you would get from a basic select query. 

**I would probably consider a set to be a Record analogous to a row in SQL**
I'm not so sure about using the word Record. A lot of PHP frameworks use `ActiveRecord` to describe Model type classes. So I always find it confusing.
What I did here was make the "row" be a `Collection` and multiple rows just be multiple `Collection` inside an array.

**Since id and probably label are so common, would we want to duplicate them at the top-level of the set? $consistent->id; as well as consistent->meta()->id. This can be done easily enough with magic methods**
This all boils down to our stance on using either id or label as vertex properties. I feel that id is safe. Not so sure about label though. 
I do think it would make our lives easier (at least in the beginning) if we considered those two to be reserved words and duplicated them on the top level.

**Also, what is the difference between `label` and `type`**
`type` can be one of three things if I'm not mistaken: `vertex`, `edge`, `property`. This piece of information could potentially be crucial to the Models mapping the data afterwards.
`label` is like you already know a way of grouping a set of vertices or edges. Though we should not consider this as a piece of schema information as some DBs don't support them (mostly older versions though). We should treat it like a higher level property.

**What is immutable in sets? Do we care about protecting them?**
This is very database specific. Some DBs don't allow you to set your own `id` for example. I think this is best left open. If someone forces an `id`and the DB disagrees it'll let them know.
Letting users have the option of changing it although a bit of a risk does have the advantage of allowing them to -for example- replace a vertex by another one. Or is handy in some cloning scenarios.

**PSR and formatting**
Yeah sorry I wasn't sure what standard was being used. I figured it could be handled via scrutinizer or some other way with Travis so I didn't give it too much thought.

**Should we add helper methods to Collection like pop(), sort() + Arrayable functionality**
We might want `sort()` to do something else in some of the extended classes. Maybe this is best left for later when we have a clearer idea of how our models and query builders behave?

**Gremlin and Tree**
Yes gremlin supports tree but there's currently a serializing issue with Gremlin3 and gremlin-server that generates an error when using this functionality. You can follow that issue here : https://issues.apache.org/jira/browse/TINKERPOP3-732
Orient could potentially use gremlin to accomplish this task. `tree()` is fully functional in TP2.

We can always just lay the ground work. Put the methods up and just have them throw a `NotSupportedException` or something.

electricjones:
All right, almost there:
- Refactor to `formatAsX`
- Let's just use a general `FormattingException` for now if the driver encounters any errors in formatting a response to a whatever. If we can find a better way to be more specific later, all the better
- If the `formatAsX` succeeds, lets add a property on the returned response, `$format` that simply labels what kind of format it is. When the driver sends back the initial response, the `$format` can be set to `"response"` or something. That way we will have a basic way to do `$response->getFormat()` and have an idea about what we are dealing with
- Drop the word `Record` everywhere. Be consistent. I just wanted to let you know where I was coming from.
- Move label and id to top level and have the driver set them when mapping response. `meta()` will hold the raw metadata. For instance, an OrientID is actually an object with a cluster and position, and labels are `classes`. So, anyone can do `$response->id` and `$response->label` and get expected result. But, if you know what datastore you are using, you can also get `$response->meta()->class` or `$response->meta()->rid` for Orient, for example
- In any response, lets protect the `meta()` so users cannot change them in the response, and also the top level `id` and `label`. Any properties on the response are freely mutable.
- Don't worry about formatting. Taken care of by release/build process
- Let's shelve helper methods on collections for now. I am adding the trait which we can use if we want. If any conflicts do exist, its easy enough to manage from traits with `insteadof` operators. But, lets see what would be helpful first.
- I will look into Tree once we settle on a format. If Gremlin already supports it, should we just use that format?

What is the best way for me to add my updates to `OrientDriver` and such. Should I do a PR against your branch, or merge and then update the `develop`. I think, I'd rather PR against yours so when we merge we merge functional code.

dmill-bz:
sounds all good to me. Lets go with the PR against my branch so that we merge functional code.
I'll go ahead and make those changes.

There's just this point to review:

_If the formatAsX succeeds, lets add a property on the returned response, $format that simply labels what kind of format it is. When the driver sends back the initial response, the $format can be set to "response" or something. That way we will have a basic way to do $response->getFormat() and have an idea about what we are dealing with_

`formatAsX` doesn't return an object (it's the driver that sends a `Response`). Instead it sends a consistently formatted array so we can't really flag it with it's type. 
I don't think this point is all that important since devs will be explicit about the result type they are expecting when writing the query. I can't think of a use case where they would need to check what type the response is. 
And from the proposals on the Query building I don't think we would need to check either. 
Should we leave this alone for now?

dmill-bz:
Oh also. Does the manager currently have any functionality around read-only properties? Maybe we should duplicate id and label to the top level and then handle the protection via the manager functionality

electricjones:
All right, we can shelve it. I can add protection of read-only functionality. It's been on my todo list for a while and shouldn't take to long. Maybe something like:

``` php
new Manager([
  'some' => [
    'data' => [
        'here' => 'value'
    ]
  ]
]);

$manager->protect('some.data.here');
//or
$manager->protect('some.data'); // protects all further nests

// So now
$manager->set('some.data.here', 'new-value');
// Would throw a ModifyingProtectedValueException
```

dmill-bz:
That sounds perfect!

dmill-bz:
Ok I made changes to this branch.

Renamed `FormatTo` to `FormatAs`

I added constants in `Spider\Drivers\AbstractDriver` for format types. The gremlin driver is capable of telling appart the various response formats without conflict (`set`, `path`, `scalar`, `custom`) not sure if other drivers will be able to.

Added `Spider\Exceptions\FormattingException`. The gremlin-server driver throws this if you try to `formatAsX` an incorrectly formatted response.

Composer update for the data-manager

Moved `id` and `label` up and protected them along side `meta`

electricjones:
Awesome. I am updating the OrientDriver now. I am having a problem with the fact that OrientDB returns an interger with the rows affected on a delete operation and it seems that GremlinServer returns an empty array? So, I'm just manually checking for that for now.

Once this is done, we should be good to merge, then clean up the develop branch and work on some of the other minor refactors.

We should create a base DriverTest that others extend from. The `SpecificDriverTest` could have various methods to fetch the queryScript and expectedResults for the different operations. The assertions would remain the same for all drivers.

If you find yourself with some time, I have started a spider/docs repo with markdown (mkdocs) files for the documentation. Some documentation on [response formats](https://github.com/spider/docs/blob/master/docs/responses.md) would be helpful. I can take care of the rest.

dmill-bz:
Ok I'll go ahead and do that. 
And I agree about the testing. I had added that to issue #13 . I think we need both unit tests for each driver and one functional `DriverTest` that we use to make sure all Drivers are consistent. Ideally we could automate it to run against every driver since it shouldn't have to be extended for each case (though this would require a functional command builder)

chrismichaels84:
I had pictured that as an additional layer, integration test suite: testing from CommandBuilder to response

The issue is that all machines would have to have all datastores setup with identical test databases. Just a hazard of the game, i guess.

dmill-bz:
Documentation done.

dmill-bz:
Ok I merged everything. I just updated the OrientDriver with the new Exception classes I created (mostly `NotSupportedException`)

If all's good with you we can merge this.

electricjones:
"If code were poetry, this would be a sonnet." Merged.

Issue #12, opened 2015-07-21
getters, setters and Meta Properties


dmill-bz:
recap : 

---

Also another issue I noticed is that graph elements have metaProperties and properties. 
Meta properties are things like `id`, `label`, `type`. Properties on the other hand can be anything the user sets (including one of those three). With the current structure where SpiderGraph is populated with a combination of both this can lead to conflicts (if say the vertex has a `type` property for example).

Consider the following to see where conflicts could arise (you can see that properties is set in SpiderRecord to avoid said conflicts):

``` php
    /**
     * Hydrate a SpiderRecord from a gremlin-server response array
     */
    protected function arrayToSpiderRecord($array)
    {
        // Or we map a single record to a Spider Record
        $spiderRecord = new SpiderRecord();
        $properties = [];
        foreach($array['properties'] as $key => $value)
        {
            $properties[$key] = $value[0]['value'];
        }

        $spiderRecord->add([
            'id' => $array['id'],
            'label' => $array['label'],
            'properties' => $properties,
        ]);
        return $spiderRecord;
    }
```

---

**Meta Properties**

Completely agree. Maybe records should be mapped in such a way that every result has a meta bag with meta properties?

``` php
$result = query->one();
$result->username; // normal property for username
$result->birthday; // normal property birthday
$result->label; // normal property for label, maybe like a tag system or whatever
$result->meta->label; // meta property for class or type, automatically populated
```

While on that topic. We should come up with a standard vocabulary. Different databases have different terms for label, class, property, field, etc. May be best just to stick with something like Neo4j or Gremlin since they are the most used.

dmill-bz:
I think we should base ourselves on TinkerPop here as it's a collection of all possible combinations. 

A few things that might need thinking if we go the `meta` route:
- is `$record->meta->id` acceptable to get the record id? Seems weird.
- how does one extend `Record` to make a `UserRecord` with a `label = "user"`
- not all databases accept custom id. Do we need to make sure users don't override this property?

Some getter/setter thoughts:

``` php
$record = new Record(); // empty record
$record->meta->label = "user";
$record->name = "john"; //does this work? What are the dangers?
$record->age = 25;
$record->getAge(); // returns 25, does this work out of the box?
$record->save(); // does a $record->insert() since no ID is set.
$record->meta->id; // some id

$newRecord = Record::select()->where('name', 'john')->one();
$newRecord->name; // john
$newRecord->age; // 25
$newRecord->meta->id; // some id (same as previous)
```

[...]

PR #25, opened 2015-07-29
CRUD CommandBuilder and Orient CommandProcessor

electricjones:
Finishes the basic CRUD command builder (no traversals) with an OrientDB command processor with tests. Documentation is next on my list.

Open to discussion.


electricjones:
A couple things to discuss about the CommandBuilder at this stage:

**How do we want to handle telling the datastore what to send back?**
For instance, in OrientDB, after a DELETE command, you have the option of returning the number of records deleted, or the full record before it was deleted. On UPDATE, you may request the record before, after, or a number of affected records. I believe this choice should be passed on to the user, but with defaults in place.

Right now there exists `$builder->insert($data)->into('target')->fromDb()`
1. I don't like `fromDb()`, but `return()` is reserved and `after()` or `then()` seem too much like callbacks
2. How to standardize accepted parameters to abstract away db-specifics?

dmill-bz:
I'm reading through the code as we speak.
Here's kind of how I saw things with the `Builder`. To me this layer is only a string manipulation library. It shouldn't be aware of either `Connection` or `Driver` or any kind of response from the DB. It simply holds the configuration for a query and converts it to a language when requested via a method -- like say `toLanguage()` -- (with extra functionality around binds)

I think the Builder will be the most popular part of the spider library as it could easily be used inside existing code and drivers.

With that in mind. A `Query` class that would extend `Builder` and be used in the context of Models is a different matter.
`Query::delete()` or `Model::delete()` can programmatically ensure that the output is the expected one. I think delete is a perfect (and perhaps the only) situation where we would need to force consistency, but it should be done on that level imo. In this case it would be as simple as just ignoring what the DB returns at all and returning an empty set.

dmill-bz:
ah my bad update is a very good point. Is there any case where you wouldn't expect an entire set of updated elements in the response?

electricjones:
As it stands, the Builder only has a dependency for a connection and a processor. The builder does nothing more than build up a CommandBag and shove it into a processor (dependency) before shoving it through a connection (dependency) and then back through a driver's response formats before returning a response.

Are you saying you would like to split the current builder into an
`AbstractCommandBuilder` which only builds up a command bag and then a `ConcreteCommandBuilder` or whatever which extends and implements the processing and dispatching through dependencies?

dmill-bz:
Ok So first off: nice job, this looks great.
Here are a few things to look at:
- [x] CommandProcessors should probably be in their own Language folder. Such as `Spider\Commands\Languages\OrientSQL\CommandProcessor`
- [x] Easy fix :  in your orient `CommandProcessor`, you return a `Command` without passing it the language param
- [x] It might be worth introducing `BaseBuilder` that holds all the basic methods like `where()`, `limit()`, `offset()`, etc.. And then have `Builder` hold the syntaxic sugar methods like `orWhere()`, `andWhere()`, `first()`, `record($id)` (which is essentially a where('id', '=', $id)), etc.
- [x] I would remove any reference to `Connection` or any DB calls (essentially I would remove `dispatch()`). Reasons for this are:
  - Like I said earlier I think `Builder` should be a string manipulation library.
  - As a user you might not want to execute the command right away. Maybe you want to hold onto it and apply scopes later on.
- [x] Introduce a `toLanguage()` method that would fetch the appropriate `commandProcessor` and return a `Command`.
  This would allow us to do either of these:
  
  ``` php
  $builder = new Builder();
  $builder->select()->...->all();
  $driver->executeReadCommand($builder);
  // or
  $driver->executeReadCommand($builder->toLanguage('cypher'));
  ```
  
  I think this is better than passing an obscure `Bag` to the driver. Though this point feels a tad weird because "fetching" the `commandProcessor` isn't as straightforward as it sounds.. Worth discussing.
  Perhaps, call it a `process()` and give it a command processor .

dmill-bz:
To reply to your question I think `Builder` (or whatever we decide to call it in the end) should simply be a string manipulation class. 
Lets take the following example: You're a user with an existing project. You pass hard coded orient queries to your orient driver and this has several pitfalls you've been struggling with (like not being able to use scopes properly, etc...) Maybe you've even been mixing gremlin in there (which means learning two languages). 

Spider comes along and you realize that you don't need to refactor you connection class or your driver classes. You can simply take advantage of query building on it's own and start refactoring your hard coded queries into Spider queries that will then be translated to whatever language your driver expects.

The requirements for this is for `Builder` to simply output a string. Regardless of Connections and dispatching. etc.

I don't know if I'm making sense.
`Query` would be the spider class that would integrate the `Builder` with the `Connection`

electricjones:
Agree to splitting the processor from the driver. My only concern with that is simplicity at the top level. The repo should be usable in its component parts, but I image most people would like to populate configuration data and then work through a single object or create Models that extend a `Spider\Model`. Especially if one of the main goals is an easy learning curve from MySql.

Happy to split into BaseBuilder and Builder.

However, I feel that the LanguageProcessor like OrientProcessor, should receive a consistent Bag with consistent properties and then create a `Command` from there, which would be executable by the driver.

Honestly, I hadn't considered existing project workflow. I am working on this for a new project. That's great.

How about this. 

So, in long hand

``` php
// Setup a command bag
$builder = new Builder(); // no dependencies other than Bag, which it creates
$builder->select()->where()->all(); // no limit

// Choose your language processor
$processor = new GremlinLanguageProcesor();

// Execute the request
$driver = new GremlinDriver($configs);
$response = $driver->executeReadCommand(
   $processor->process(
       $builder->getCommandBag()
   );
);

$set = $response->toSet();
```

Or, from use within an existing project, if all you want is the Command

``` php
$builder = new Builder(); // no dependencies
$language = new GremlinLanguageProcessor();

$query= $query->select()->where();
$gremlinQueryScript = $language->process($query)->getScript();
// We could type-hint the language processor so you can pass a Builder
```

Or, from a higher abstraction

``` php
$query = new Query(new Connection($driver), new LanguageProcessor()); // extends Builder
$set = $query->select()->where()->set()->all();
```

Or from an even higher abstraction (to be implemented later)

``` php
Spider::configure([
   'default' => 'my-connection'
   'my-connection' => [
       driver => 'gremlin' // which sets the driver and language processor
       username => 'whatever' // and others
   ];
])

$spider = Spider::make(); // extends Query
$result = $spider->select()->where()->all(); // returns set by default
// or
$result = $spider->select()-where()->path()->get(); // returns a path
```

electricjones:
The driver never receives anything except a `Command` which is just a wrapper for the native script. The driver never receives a `CommandBag`

dmill-bz:
> Agree to splitting the processor from the driver. My only concern with that is simplicity at the top level. The repo should be usable in its component parts, but I image most people would like to populate configuration data and then work through a single object or create Models that extend a Spider\Model. Especially if one of the main goals is an easy learning curve from MySql.

This is what the `Query` Would do. We would promote `Query` as the actually Spider command builder but users will have the option of falling back to `Builder` if they have specific needs.

> However, I feel that the LanguageProcessor like OrientProcessor, should receive a consistent Bag with consistent properties and then create a Command from there, which would be executable by the driver.

I absolutely agree. I like how it currently works, I was just bringing up a point that came up in one of earlier conversations which was that we could pass a `Bag` to the driver, which I don't think we should do.

> So, in long hand

Exactly. We can either pass the Builder to a processor or a processor to the Builder. I feel that the later would be the better option because it would allow for : 

``` php
// Setup a command bag
$builder = new Builder(); // no dependencies other than Bag, which it creates
$builder->select()->where()->all(); // no limit

// Choose your language processor
$processor = new GremlinLanguageProcesor();

// Execute the request
$driver = new GremlinDriver($configs);
$response = $driver->executeReadCommand(
       $builder->process($processor)
);

$set = $response->toSet();
```

As well as : 

``` php
$driver = new GremlinDriver($configs);
$response = $driver->executeReadCommand(
       $builder
);
```

The difference with the later is that the driver would have the option of choosing the processor. For example, if you're using `Spider\Drivers\Gremlin\Driver` with a Neo4J backend, the driver can decide to send your query in `cypher` instead of gremlin (this should be the default behavior).
This also has the benefits of allowing the users to be able to simply ignore the language, both in their higher level configuration or when doing lower level driver handling. 
Just like follows:

``` php
Spider::configure([
   'default' => 'my-connection'
   'connections' => [ // I don't like this term. This is a package of driver, processor, and credentials
     'gremlin' => [
        'driver' => 'Full\Class\Driver',
       // 'processor' => 'Full\Class\Processor', Ignore this if you want
        'hostname' => 'localhost' // and the rest of configurations    
     ],
   ]
])
```

Simplifies things for the user. Unless they need to they can be oblivious to how processors work.

> Or, from use within an existing project, if all you want is the Command

Just one change:

``` php
$query= $builder->select()->where()->all(); // the all() from the Builder is different from the all() of the Query in that there are no db calls etc.
```

dmill-bz:
``` php
$builder->select()->where();
```

and

``` php
$builder->select()->where()->all();
```

Are essentially the same query. But:

``` php
$query->select()->where();
```

and

``` php
$query->select()->where()->all();
```

Are different if that makes any sense.

electricjones:
Yes. I like. Refactoring now.

electricjones:
Btw, `record()`, `records()`, and `byId()` are all aliases of `target()`. They actually set the target to the record ID, not add a where constraint.

In most cases, this would equal the same thing, but in OrientDB it actually makes a difference for two reasons.
1. There actually is no `id` property. It is `@rid`.
2. `SELECT FROM #12:1` is much more performant that `SELECT FROM Users WHERE @rid = #12:1`

It would be up to the language processor to decide the best way to add that, I guess.

dmill-bz:
That's really interesting! I guess it makes sense as most of the graph DBs used to function like this (TP2 as well). It might change when orient implements TP3 but for the time being this can work just fine.

electricjones:
I have pushed a rough commit that splits the builder into:
- `BaseBuilder` with basic CRUD capabilities. It creates command bags, and nothing else
- `Builder`: which extends the base and adds sugar
- `Query`: which holds a connection and processor and adds more sugar to Builder

Updating the tests now, to make sure nothing exploded, lol.

dmill-bz:
Looks great.

What are we going to do about `all()`, `tree()`, `path()`, `scalar()` (and potentially `set()`)?
The builder will need to be aware of these as it isn't just a return format, it's also part of the query.

Originally my thinking was along the lines of having `all()` and `one()` return a `Set` and not serve as a `LIMIT` I'm thinking that may be confusing (or not). What do you think?

dmill-bz:
To elaborate further,  both of these would return the same thing:

``` php
$query = new Query(/* config */);
$query->select()->where()->limit(1)->all();
$query->select()->where()->one();
```

And this wouldn't be possible:

``` php
$query = new Query(/* config */);
$query->select()->where()->path()->all(); // path and all would both dispatch.
$query->select()->where()->path()->one(); // same here
```

Instead you would do:

``` php
$query = new Query(/* config */);
$query->select()->where()->path();
$query->select()->where()->limit(1)->path();
```

Does that make sense? Is it too confusing?

electricjones:
I hadn't considered the response formats to be part of the query. In my mind, you could use the `Builder` without a driver. 

If all you want is to build queries because you have your own driver in place, you can just use `Builder` and a language processor to produce a script and then pipe it through whatever driver you wanted.

The response formats are specific to spider-drivers and belong in `Query`. So, what the `set()` and so on in Query do, is simply declare that you want a set when you dispatch.

`all(),; one(); first(); get();` and `dispatch()` are the only dispatch methods. The dispatcher executes the query through the driver and simply calls the `formatAsX` method. `formatAsSet()` is default.

You want to be able to dispatch from `tree()`, `path()`, `set()`, and `scalar()`?

dmill-bz:
Well `tree()` and `path()` are going to be required on the `Builder` level. Here are a couple of examples:

``` php
$builder = new Builder();
echo $builder->select()->where('name', 'marko')->process($processor); // all() on this level is like not using anything. But we could ->all() here
echo $builder->select()->where('name', 'marko')->path()->process($processor);
echo $builder->select()->where('name', 'marko')->tree()->process($processor);
```

in `gremlin`:

```
g.V().has('name', 'marko');
g.V().has('name', 'marko').path();
g.V().has('name', 'marko').tree();
```

In `cypher`:

```
MATCH (n {name:'marko'}) RETURN n;
MATCH p=((n {name:'marko'})) RETURN p;
//Still need to figure tree out for cypher but it will be different.
```

So from the above we can see that `tree()` and `path()` are required by the `Builder`.

On the `Builder` level:
- `all()` would essentially do nothing (which might be confusing because you could expect it to remove any `LIMIT`)
- `one()` would add a `LIMIT 1` which is redundant with `limit(1)` and `first()`
- `scalar()` just like all wouldn't do anything.
- `tree()` does it's job of flagging the Bag for tree
- `path()` does it's job of flagging the Bag for path

So all things considered, I would be for removing `one()`, `all()` and `scalar()` from the builder, but add `tree()` and `path()`.

On the `Query` level:
- `all()` would set the format to `Set` and dispatch.
- `one()` would apply `limit(1)`, set the format to `Set` and dispatch
- `path()` would run the `Builder::path()`, set the format to `Path` and dispatch
- `tree()` would run the `Builder::tree()`, set the format to `Tree` and dispatch
- `scalar()` would set the format to `Scalar` and dispatch

electricjones:
Sounds great. I just pushed a bunch of working tests and refactors.

There are a lot of duplicated tests here, but we can weed those out later.Now, all passes.

Still todo:
- [x] Refactor `Builder` and `Query` to use all, one, path, tree, and scalar correctly
- [x] Implement `toLanguage()` where did you see this living?
- [ ] Add documentation

Anything else?

I'm going to be gone for a while, so I wanted to get this up for discussion/refactoring.

dmill-bz:
`toLanguage()` would be the `process()` that was discussed earlier. It's essentially a `Builder::getScript()` but with a processor param. I'm not sure what the best naming policy would be, we already have three options here. Maybe just `getScript($processor)` would work.

I think that's about it. It's looking good.

electricjones:
All right, I think this is all the functionality we've talked about:
- `Builder::getScript()` passes the current `CommandBag` through a language processor. You may set the processor via method, constructor, or pass it through `getScript($processor)`
- `Builder::first()` sets limit to 1
- `Builder::tree()` flags command bag as tree
- `Builder::path()` flags command bag as path
- No all, one, set, scalar on Builder.
- `Query` now sets flags, dispatches command, and formats via:
  - `Query::set()`, `path()`, `tree()`, `scalar()`
  - `Query::all()`, `first()`, `one()` returns formatted as a set
  - `Query::dispatch()` returns the `Response` sent back by the `Driver`
  - `Query::command($command)` allows you to pass native string scripts or `Command` objects

Revised tests, though they need some love in general.

Also, note that `Command` now has a `getRw()` and `setRw()` method to flag the Command as a Read or Write Command (so `Query` can pass the command to the correct diver method)

electricjones:
Documenting it all now :+1:

electricjones:
Documentation has been added at http://spider-ogm.readthedocs.org/en/latest/command-builder/.

I think this is ready to merge.

The nice people over at readthedocs have helped me solve my weird repo problem, so (after this merge), I'll move /docs/ into this repo.

**Also note**
- `DriverInterface` now has a `makeProcessor()` method that returns the default desired language processor.
- `Query` only requires a `Connection`. The default processor is taken from `DriverInterface`. Of course, you can override this with a second constructor param
- Gremlin and Neo4j have been updated, but only throw a NotSupportedException, since they have no language processors yet.

dmill-bz:
Looks great. I have a couple of questions.
- Should the `Bag` be aware of the `FORMAT`? Right now it seems like `Query` passes the format along to `Bag` who then passes it back to `Query::dispatch()`. Did you have another use case in mind? Couldn't we do, say:
  
  ``` php
  class Query extends Builder{
    public function all(){
       //.. stuff happens
       return $this->dispatch()->getSet();
    }
    public function tree(){
       //.. stuff happens
       return $this->dispatch()->getTree();
    }
  }
  ```
  
  This way we're using the `Response` interface and not the `Driver` directly and there's no need for passing along the format to `Bag` (which just helps manipulate the query string and doesn't really care about `Response`s anyways).
- How would the `DriverInterface::makeProcessor()` work in events where drivers choose between multiple processors? `GremlinDriver` would need to either use a `GremlinProcessor` and send the query as is. Or use a `CypherProcessor` and send `graph.cypher(<cypher query here>);` (which I guess still counts as `gremlin`?)

electricjones:
You had said that you wanted the CommandBag to be flagged for tree and path. That's all it does is flag the CommandBag so the command processor can know if it needs to factor that into building the query script.

The Query::$format property is what decides what Response Format is returned.

`Driver::makeProcessor()` is never required. It's just a way to avoid having to pass two arguments to the `Query` constructor. `Query` needs a language processor, and this way the driver can choose which it wants, unless the user gives `Query` a specific language processor.

We may not be understanding each other.

dmill-bz:
Ohh I see.  So we're understanding each other it's just the naming and the way dispatch works that's a little non-intuitive. 
It's true that we're flagging a `Bag` to build a query that will return a `path` or `tree` but technically (and this is just semantics) a `Bag` shouldn't understand what a response format is.
I feel like theres a need to clearly separate these `FORMAT`s from the `Response` formats you would expect from `Query`.

Even though it might seem weird I think the logic between the `Builder` and the `Query` should allow for something along the lines of this to be possible:

``` php
class Query extends Builder
{
   public function newType() // a new type like all(), one(), tree(), etc..
   {
      // ... code here
      $this->bag->format = Bag::FORMAT_SOMETHING;
      return $this->dispatch()->getSet();
   }
}
```

The above may or may not throw an error depending on wether `FORMAT_SOMETHING` is compatible with `getSet();` but it should be possible to do it. 
Right now the logic is too tied together to allow for this.

Because of this I would be tempted to rename `Bag::format` and `FORMAT_X` to something else. I'm not sure what though. Maybe `MAP_X`? Let me know what you think.

And have `dispatch()` return a `Response` that is then formatted via `getX()` by the calling method.

As far as the `Driver` goes. I'm good with `makeProcess()`

dmill-bz:
PS: I realize that in my example `$this->bag->format = Bag::FORMAT_SOMETHING;` should actually be in the `Builder`. But I wanted to keep it simple.

electricjones:
Alright, I get you. So, lets say `Bag` has a flag for `map` with `Bag::MAP_TREE`, `Bag::MAP_PATH`, and maybe `Bag::MAP_SET` by default (no mapping).

`Query::format` will flag the query builder to return whatever response format is requested (tree, path, set, scalar, or custom).

`Builder` should have no response format. It doesn't format responses, only flag command bag for mapping.

I'm not sure I follow with `newType()`. Would you mind if I merged this, and let you simply work that directly into the `develop` branch. It seems like a minor change.

You have committing rights, yes?

dmill-bz:
Yes. No problem I can make these changes. The PR is great as is, the rest is really minor.

Issue #27, opened 2015-07-31
Test Restructure


electricjones:
- [x] **Restructure Tests**:
  Testing should make use of proper Mocks instead of hashing together so many spaghetti stubs. Tests should also be reorganized so that Mocks and Stubs live alongside the tests that use them in a more modular way.

``` bash
/tests/Unit/Connections/ConnectionTest.php
/tests/Unit/Connections/ConnectionStub.php
```
- [x] **Test Coverage**: Test coverage is pitiful, but I bet its not as bad as it looks. Just needs a once-through

dmill-bz:
This is slightly off topic although still pretty relevant. I Think we should provide some `Fixture` functionality and use this for our testing. This goes for #26 as well.

It would definitely be useful for spider users. The question is, how do we implement a basic graph array?  Should we first concentrate on how we would structure a `Response` as a `graph` or `subgraph` (is it even possible to consistently provide this functionality over various drivers)? Then use that format?

dmill-bz:
Ok after looking at this a little closer here's what I think we could do:
- Have a single `DriverStub`. This might be a class that has to do a bit of work. I think the idea is that we should decide on a bogus `_raw` format and then build correct `formatAsX` methods to map that `raw` format.
  Once this is done we should be able to modify the behavior of `executeReadCommand()` and `executeWriteCommand()` on a per test (or TestCase) basis via phpunit's `$this->getMockBuilder()`
- A `ProcessorStub` that returns the `Bag` object rather than a `Command`. We can use this to test that `Query` Methods affect the `Bag` correctly (`Query::all()` etc.). As well as testing the `Builder` methods to some extent. Though again this we could do with  `$this->getMockBuilder()` but I feel it's going to be so often that it warrants it's own stub.

I think with those we can pretty much do what we need to do. Does it feel like we're missing something?
- `Driver` testing doesn't really needs any stubs since we want to test against DBs.
- `Response` testing is Driver dependent, it would need the `DriverStub` mentioned earlier.
- `Query` and `Builder` testing would need a `ProcessorStub` as mentioned above
- `Connection` testing ?
- `Processor` testing is going to be on a per language basis and is going to be dependent on `Builder` so no need for stubs here?

This will already be a significant clean. What do you think?

electricjones:
This has all been on my todo list, but you're right. It needs to happen sooner rather than later.

For now, `Connection` won't need any stubs. It's only real work is passing call through to the driver.

[...]

PR #30, opened 2015-08-03
Made a quick fix to how Queries return their data and Changed naming …

dmill-bz:
…of Bag::FORMAT_X to Bag::MAP_X

### DIFF
--- a/src/Commands/Bag.php
+++ b/src/Commands/Bag.php
@@ -75,13 +75,13 @@ class Bag extends Object
     public $orderAsc = true;
 
     /**
-     * Flag a response format for the query to return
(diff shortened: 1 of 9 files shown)


dmill-bz:
Turns out I got deeper and deeper into the test stubs until I decided to backtrack and keep this PR ultra simple. 
I've had to disable Query tests though as they didn't pass (they were expecting incorrect return values anyways). I'll elaborate more on this in #27 .
I think the next step is to really get the testing restructure in place. The more we postpone the harder it's going to get.

All in all this is a very simple PR, shouldn't be too hard to merge this in. The testing situation is another matter.

PR #32, opened 2015-08-05
Made some test refactoring around issue #27 . Removed most stubs

dmill-bz:
I've removed most stubs. All tests pass.

I've however left `CommandProcessorStub` which I believe we can remove. Removing it however will depend on a readjusting of responsibilities in regards to the command processing which I feel is a little fuzzy right now. (This is a little related to our discussion around Languages in PR #25)
Since I didn't want to add that to this PR i've left the stub there. 

I also added a public `BaseBuilder::getBag()` this simplified the testing but if you feel it shouldn't be available in the API I can fix it pretty easily. (I just thought it wasn't hurting)


dmill-bz:
To elaborate on the "fuzzy" comment. The question with no clear answer at the moment is **Who's job is it to provide the processor, and who's job is it to process the Bags?**
Part of it is done in the configuration if you want to specify a language. `Query` processes the bags before dispatching to the `Driver`. `Driver` provides a `Processor` as well. It's a bit of a mess.

We should probably open an issue around this.

electricjones:
Seems much cleaner and I appreciate that. The spegetti stubs grew up because I didn't have a clear sense of direction at the outset, and they just got out of control, lol.

Just FYI, `BaseBuilder` already had a `getCommandBag()` which returns the bag. I like `getBag()` better, so we can just drop its alias.

A couple questions/concerns
1. This is probably related to my next two comments. This PR removes any testing of `dispatching` from the `Query`. Is this something you don't feel is important to test? Or is it just that we need to find a better way to test it? I feel like we need to test that when you dispatch from `Query` it:
   - sends the correct bag through the processor
   - sends the response through the right `formatAsX` method.
     Which is why I was originally just json_encoding the Command and dynamically flagging it with the requested response format. I'm stumped over a better way to do this.
2. I don't fully understand what you mean by fuzzy responsibilities. Only the processor ever processes bags. The processor is an optional dependency in `Builder` and a required dependency in `Query`, but neither of those actually process anything. They just hand it off to the processor and use the resulting command. This doesn't seem like shared responsibility to me. 
3. The question of who supplies the processor is an issue I've struggled with. I think the problem is that we are in a WIP stage and there is no global configuration yet. I had originally thought that only the driver should provide its preferred processor, but you had wanted the option to override that, unless I misunderstood you. I can see why that would be nice: if you are dispatching gremlin directly to Neo4j without using gremlin server, for instance. But, I assume, that would require the driver to dispatch differently (in orient it would). To that end, we created the option in `Query` to pass it via configuration **only** if you want to override the driver's preferred processor. Even that wouldn't work in the current drivers because they don't check the script language from the `Command`, though it would be easy enough to.

In good OOP style, the answer maybe another class `ProcessorManager` or something.

dmill-bz:
Ah thanks I completely missed the `getCommandBag()`.
1. Yes you're correct. There are kinda two reasons here:
   - First one is that depending on 2. & 3. the expected input for `Driver::executeReadCommand()` might change so I felt like it was more appropriate to write the tests after.
   - Secondly,  The chain gets broken in the driver query methods. At this stage, from type casting alone we know that the dispatch is issuing the correct type of element to the driver. Whether the content of this element is correctly formed is something that can be left to `CommandProcessor` tests. And whether or not the Driver returns the correct information should be left to `Driver` tests.
   
   So in essence I guess that my feeling is that testing the full chain is more of a functional test scenario than a unit one. 
2. (+ 3) Well for example, should it be a required dependency for the `Query`? After all we should probably delegate the choice to the `Driver` entirely while allowing forcing of the language (within the `Driver`'s scope of course). For example in pseudo code:
   
   ``` php
   class Driver extends AbstractDriver implements DriverInterface
   {
     protected $languages = [
        'gremlin',  // default
        'cypher'
     ];
   
     public function executeReadCommand($query)
     {
        $processor = $this->getProcessor(0); // set default processor
        if ($query instanceof BaseBuilder) {
           if ($this->isNeoBackend()) {
              $processor = $this->getProcessor(1);
           }
           $query = $query->getCommand($processor);
        }
        // else assume Command. either check or not
        if(!$this->isSupportedLanguage($query->language)) {
           throw new NotSupportedException(__CLASS__ . " does not support ". $query->language);
        }
   
        // if you need to take care of binds
        if($query->language == 'cypher') {
           $processor->replaceBinds($query); // will also remove $binds from the Command.
        }
   
        //submit to server by whatever channels are required based on language.
     }
   }
   ```
   
   I feel like `getScript()` should just return a string (for integration into other systems). `getCommand()` could return the command.
   If the Builder already has an assigned processor (from Spider configuration for example) then it would pass a `Command` to the driver.

Does this make sense?

electricjones:
So, in this case, the `Driver` also accepts any of the Builder instances? 

And `BaseBuilder` (or more likely `Builder`) would have a `getCommand($processor)` method which runs the current `Bag` through the processor its handed?

By default, `Query` and `Builder` pass themselves to the driver,  which uses its default desired processor to create a `Command`. 

If configuration at the `Query` level specifies a language processor `Query` passes a `Command` to the `Driver` which reads the `getScriptLanguage()` and executes the command appropriately.

I don't think `Builder` should be able to interact with the `Driver` at all. It's sole purpose is building command bags. I'm fine with a one-off method for processing that `Bag`, but the processor MUST be handed to `getCommand()`. The rest of `Builder` is entirely driver/processor ignorant.

`Query` is aware of a `Connection` and therefore a `Driver`, so it can dispatch itself.

Am I understanding?

dmill-bz:
**So, in this case, the Driver also accepts any of the Builder instances?**
Yes.

**And `BaseBuilder` (or more likely `Builder`) would have a `getCommand($processor)` method which runs the current `Bag` through the processor its handed?**
I think `BaseBuilder` would probably be the best place. And that is correct. Essentially It just gives two output options for the builder. Either in `Command` form on in `string` form. 

**By default, `Query` and `Builder` pass themselves to the driver,  which uses its default desired processor to create a `Command`.**
Only `Query` would. `Builder` doesn't know about the driver. It just has an extra new way of outputting scripts. `string getScript($processor)` and `Command getCommand($processor)`

If we're ok on this and on point 1. Then maybe we can merge this and I can make a new PR with these changes.

electricjones:
Yes, this PR is great.

I'm just not crazy about `Driver` accepting anything other that `Commands`. To me, the further down we go, the more important it is to follow the single responsibility mantra. A driver should only abstract a datastore's native functionality -- primarily dispatching commands. 

I wasn't even crazy about the `formatAs` methods being part of the driver, but there is simply nowhere else for them to live.

I do like this proposal, I'm just concerned about `Driver` bloat. Let's move forward now. We can always split `Driver` later if it becomes too bloated.

I'd also like to start declaring a class's primary responsibility in the class's docbloc. Just as a guide to keep things slim.

dmill-bz:
Yeah I hear you. Maybe there's another way. It's a little tricky because only the driver will know the best way of handling binds and that involves implementing all the Language processor portion of the bloating.

In contrast, the passing of a Builder is a single if (so no real bloating involved here).

I thought of implementing some sort of callback that `Query` could use to figure out how to format the `Command`. But the problem with that is if you're using the `Driver` on it's own, then `Command('script', 'language', $binds)` would not be cross-driver compatible. `GremlinDriver` would function properly whereas `Neo4JDriver` wouldn't know what to do with $binds. 

I'm at a loss for a better option.

dmill-bz:
Anyways, this PR doesn't really impact on any of that so I'll merge it and we can see if anything can be done about the driver situation in the next PR (or if we decide to even adopt it).
It might make more sense when we have code to look at.

Issue #26, opened 2015-07-31
Consistency across driver testing and fixtures


electricjones:
Now that we have multiple drivers, it would be nice to have a single `DriverTestBase` that other drivers would extends to ensure all functionality is test and tested consistently. It would also give a single point to modify most aspects of the test when new features are introduced or bugs are squashed.

My thought was something like this. Each driver would extend, implement, and setup their specific driver. They may have to do a little more once we go through the tests to see what can be abstracted and what needs to be uber driver-specific.

``` php
abstract class DriversTestBase extends \PHPUnit_Framework_TestCase
{
    use Specify;

    protected $config;
    protected $driver;

    /* MUST IMPLEMENT */
    /* MUST ALSO USE setup() TO CREATE $this->driver AND $this->config*/
    public $queries = [
        //'alias' => [
        //    'command' => 'COMMAND OBJECT',
        //    'response' => 'RESPONSE OBJECT TO TEST AGAINST'
        //],
        'read-one-item' => [
             'command' => new Command("SELECT FROM users LIMIT 1"),
             'response' => [
                'id' => '#9:1',
                'name' => 'HEY BO DIDDLEY',
                'label' => 'V'
             ]
        ],
    ];

    public function getCommand($which)
    {
        return $this->queries[$which]['command'];
    }

    public function getResponse($which)
    {
        return $this->queries[$which]['response'];
    }

    public function testReadCommands()
    {
        $this->specify("it selects a single record and returns an array of Records", function () {
            $this->driver->open();

            $response = $driver->executeReadCommand($this->getCommand('read-one-item'));

            $driver->close();

            $expected = $this->getResponse('read-one-item');

            $this->assertInstanceOf('Spider\Drivers\Response', $response, 'failed to return a Response Object');
            $response = $response->getSet();
            $this->assertInstanceOf('Spider\Base\Collection', $response, 'failed to return a Record');
            $this->assertEquals($expected['name'], $response->name, "failed to return the correct names");
            $this->assertEquals($expected['label'], $response->label, "failed to return the correct label");
            $this->assertEquals($expected['id'], $response->id, "failed to return the correct id");
        });
    }
}
```

dmill-bz:
I like this. It's simple and does what it should

electricjones:
I already have a local branch working on this, so I'll get it up when I can. I won't have as much time this week to work on this. Other deadlines are approaching :(

dmill-bz:
Yeah I know the feeling! Good luck with it all.

[...]

PR #33, opened 2015-08-05
Refactor driver tests around #26 - driver tests extend BaseTestSuite

electricjones:
**This is not finished**. I just wanted to get my wip up to make sure sure it was in an agreeable direction. I will refactor the Gremlin and Neo4j tests next. And I intend on using a similar process for Lanaguage Processors.

I decided to use abstract methods in the TestBase to make it easier for driver creators. Now they just implement those methods and have them return the correct command/expected pair or object. It's all noted in the docblocks. I will add a file-level docblock in the DriversTestBase class with more clarification.


electricjones:
I also wanted to say that I appreciate your help on this project and your patience. I'm still new to web development and especially to graph databases. I'm excited about the direction this is heading.

dmill-bz:
Hey it's nothing and you've done most of the work (great documentation btw). I'm also excited about the direction it's taking. I'm hoping we don't hit any major structural problems further down the road.

This is looking good. Quick question, what's the use case for the raw response methods in orientDB driver?

I'm also wondering if the expected results shouldn't be in `consistent` format, it's a little more tedious to write but at least you don't need to figure out how these arrays map to the consistent responses in the `DriverTestBase`. I'm not all that partial to either really, just a thought.

electricjones:
Where are the raw response methods?

`mapRawResponse()` turns an array of `PhpOrientRecord`s into an array of `Collection`s.

`rawResponseToArray()` ensures that the `Response::$_raw` is an array. If only a single record is returned, PhpOrient returns a `PhpOrientRecord`. This is true even if it was possible to recieve multiple records (`select from v where name = 'michael'`) could return a single record or an array of records depending on number of vetexes with name 'michael'.

It just made things easier down the chain if I could be assured that every response from the client is an array of records. I'm sure the driver could use a good refactor, but I want to wait until some of the dust has settled.

dmill-bz:
oh I was talking about `buildTwoRawResponses()`. Should this be done in GremlinDriver as well?

electricjones:
The `buildRawResponses()` methods are used in `formatAsSet()` tests.

`buildSingleRawResponse()` returns a `Response` with one record in `$_raw`.

`buildTwoRawResponses()` returns a `Response` with two records in `$_raw`.

I did this last night, and looking at it now, there has to be a better way to handle that test. 

If nothing else they should be renamed. Orient doesn't yet support path or tree, but each of those formats will also need a method to produce the appropriate raw responses for testing.

dmill-bz:
oh I see, that makes sense. 
Maybe these tests should instead involve actual DB queries and return values. This might be safer and easier to readjust if/when the DB APIs change.

electricjones:
Agreed. I'll rewrite.

After that, I can go through and adapt Gremlin and Neo4j, but I don't have the correct systems on my machine, so I can't test them. It'll mostly be copy and paste, though, so if I'm careful nothing huge should break. Just be sure to run the tests to double check.

dmill-bz:
Yeah once you feel this PR is ready to merge I'll pull it and run the tests to make sure.

electricjones:
Alright, I rebased and finished up the Gremlin and Neo4J tests.

I also refactored to get rid of the rawResponse nonsense and simplify things a bit.

The only things the BaseTestSuite don't test are `formatAsPath` and `formatAsTree`, but Gremlin and Neo4J still test these in their own test suite. Once we (or really _I_) get a handle on paths and we settle on a format for trees, we can move these tests into the BaseTestSuite.

I'm tackling a basic Processor BaseTestSuite now. It'll just have several `CommandBags` to be processed.

Please pull this to make sure I didn't destroy Gremlin or Neo tests. All Orient passes.

electricjones:
Fixed an error. Make sure to pull latest commit #80c6587

dmill-bz:
There's one issue I can tell atm. You hard code the responses sent to `formatAsScalar` but these are just as DB specific as the others. We should provide this data on a driver basis.

dmill-bz:
ie: Neo4J expects `[[5]]` instead of `[5]`

dmill-bz:
I just made a PR against this branch for the drivers
Appart from the scalar issue all other tests should pass.

dmill-bz:
I realize I might've broken orient tests on transactions.. But you probably already figured that out.

electricjones:
I fixed the orient tests, and updated the signature for `driver()` across all the drivers.

I added a `getScalarResponse($switch)` method that returns a scalar in whatever format the driver prefers. I don't see a need for an actual query for this atm.

I also finished the base test suite for the language processor tests and updated the OrientSql tests.
It works much the same way: extend the `BaseTestSuite` and implement the methods. In this case, the methods return `Command`s that are expected by the different tests. They are used as the `$expected` by phpunit.

I think that's it for this PR unless this unraveled another problem. All orient tests pass. If its good to go, go ahead and  merge.

dmill-bz:
ok let me run a final check on gremlin and neo tests

dmill-bz:
ok For some reason some of the transactional tests are failing. I need to head out now but I'll give this a proper fix tonight and then I'll merge. 
We're almost at the 0.3 mark

electricjones:
Sounds great. The Orient transactions pass, so I'm not sure what I did to break them.

We should be all good for 0.3 except a Cypher Processor and if we want to fix the processor building process we talked about. I'm good calling that a 0.4.

Also, I've installed Neo4J. Is there a way to get the database so I can run the tests locally? I assume the GremlinServer is running off the same Neo4J database? I tried to run the tests against the default database and things exploded in an epic way.

electricjones:
Well, I bet I found why the transaction tests are failing for gremlin at least.

Inside all the GremlinDriverTest queries, you need to call either a transactional or non-transactional graph. In the BaseTestSuite, we setup the correct driver, but in the GremlinDriverTest, it is using `$this->graph` and `$this->traversal` which is still pointing to the non-transactional graph.

electricjones:
Just pushed commit #b48cff7. That should at least fix that, albeit not in a terribly performant way.

dmill-bz:
ok I fixed everything. Had to change the scalar tests in the base class though so you might want to run a final check against orient.

Basically scalar tests should test against other expected result sets (Set, Multi Set, path, tree, etc..) so I made the change going in that direction other than arbitrary responses that could be anything really (ie: why test `[[10]]` and not `[[[[[[10]]]]]]` etc..) if there are really some edge cases they can probably be handled on a per driver basis. 

I'll open an issue for installing the servers. I would like to install orient as well.

electricjones:
This looks great and passes. Merged

Issue #36, opened 2015-08-09
Add integration and functional tests


electricjones:
Integration tests that test the query-builder through to the database are needed. Should test public api identically for each of the three drivers.

After integration tests, functional test that setup a full `Spider` system from the top level spider api all the way down.

I see no need for acceptance testing in spider, seeing as functional testing will test all functionality.

electricjones:
Began. See PR #43.

More tests are needed, but we don't need for them in order to merge. See the PR for needs.

electricjones:
Closing this with PR #39. More tests as we go along

Issue #35, opened 2015-08-06
Installing servers for testing


dmill-bz:
Feel free to add orient to this.
###### Download
- [Gremlin-server](https://www.apache.org/dist/incubator/tinkerpop/3.0.0-incubating/apache-gremlin-server-3.0.0-incubating-bin.zip)
- [gremlin-server-php.yaml](https://raw.githubusercontent.com/PommeVerte/gremlin-php/master/src/tests/gremlin-server-php.yaml)
- [gremlin-php-script.groovy](https://raw.githubusercontent.com/PommeVerte/gremlin-php/master/src/tests/gremlin-php-script.groovy)
- [neo4j-empty.properties](https://raw.githubusercontent.com/PommeVerte/gremlin-php/master/src/tests/neo4j-empty.properties)
###### Installation

Extract Gremlin-server and go into the created folder.

Neo4J is required for the full test suit (it serves as the transaction enabled graph). It is not bundled with gremlin-server by default so you will need to manually install it with:

``` bash
bin/gremlin-server.sh -i org.apache.tinkerpop neo4j-gremlin 3.0.0-incubating
```

Copy the following files :

``` bash
cp path/to/gremlin-server-php.yaml <gremlin-server-root-dir>/conf/
cp path/to/neo4j-empty.properties <gremlin-server-root-dir>/conf/
cp path/to/gremlin-php-script.groovy <gremlin-server-root-dir>/scripts/
```

You will then need to run gremlin-server in the following manner (it's important to run it from the `<gremlin-server-root-dir>`):

``` bash
bin/gremlin-server.sh conf/gremlin-server-php.yaml
```
###### Download
- [Neo4J Server](http://neo4j.com/download/) (Community version is tested)
###### Installation

Once the server is up and running, connect to [http://localhost:7474/browser](http://localhost:7474/browser) 
- Change the username:password from `neo4j:neo4j` to `neo4j:j4oen`
- Click on the star tab and import [this script](https://gist.githubusercontent.com/PommeVerte/46ed5dee72688ee2faf9/raw/12804b280e3e9361319afb5a4bd2084505aed1d0/Neo4J%2520modern%2520graph) that will create a "modern" graph.
###### Download
- [OrientDB Community](http://orientdb.com/download/) (Community version is tested)
###### Installation

Simply extract the directory and place it anywhere you like.

That's really it. The server can be accessed by connecting to : [http://localhost:2480/](http://localhost:2480/).

[...]

PR #39, opened 2015-08-13
Normalize testing, add basic fixtures, begin integration tests

electricjones:
I first opened the pr against the master. This is against develop.

This is finished, so long as the Gremlin tests pass. Always welcome to feedback, though.

Also see PR #43 which is a branch from this that also adds integration tests for the Query Builder. All things considered. It would be easier to merge that one and let this resolve on its own.

This pull request does several things, although maybe not in the most performant manner:

**1. Adds a mechanism to create fixtures and creates OrientDB fixture.**

Fixtures are created before any tests are run and destroyed after all tests are complete, or a fatal error is reached. It does NOT create or manage the fixture between tests.

I created a new bootstrap file that cycles through the first-party drivers and uses an instance of the `Fixture` class to setup a database for testing. This assumes that the software is installed and that the server password/username is set correctly.

ATM, only orient uses this fixture. I still don't have Gremlin setup on my computer - see #35 - and Neo4j may be a problem because it doesn't seem you can create databases on the fly. I am sure there is a solution, but I haven't dug much into it.

Each `Fixture` (`Orient`, `Gremlin`, `Neo`) includes a `setup()`, `teardown()`, and `reset()` method.

This does hit performance on the testing, though. It takes a second or two (on my computer) to create and build up the database. I don't know any way around that.

Still, now we can accurately test without worrying about state.

**2. Normalizes OrientDB** so it uses the same graph as Gremlin and Neo4j. The only subtle difference is in the ids. Orient uses a cluster:position format (#12:1) which is a string. So the ids for the vertices are different. Everything else is identical.

**3. Simplify BaseTestSuite**
Now that all drivers use an identical graph, I created an abstract of that graph in native php format to use as the expected values in phpunit. This dramatically simplified the DriverTests. Now, all they have to do is provide the commands for queries, not the expected results.

I have gone through all three drivers with the changes. Orient and Neo4j pass on my machine, and I imagine Gremlin does as well, but I haven't got it up and running yet.

Let me know if you have any comments. I am building on this to create some simple integration tests per #36


dmill-bz:
Hey I wasn't expecting this so early in the dev cycle. I've got a few things I'd like to bring up. not all of them are important. I'm short on time so rather than give this thorough thinking I figured I would make bullet points as I go and we can think about them after. Anyways this does clean up the base tests and that's cool. 
- [ ] Feels too early for fixture implementation as `Fixture` should be higher in the stack. It should depend on language processors, and should populate the graphs via `Query`. This would 
  
  a) help solve the creation of databases issue on the fly. As orient is a bit of an oddity in regards to this  (all other graphDBs function differently. Neither titan, tinkergraph, neo, sparq etc.. use the database design). It's very RDBMSy. (would imply having a test db already in place (via migration?))
  b) no need to have DB specific fixture implementations.
  c) make integration with `Model` seamless. Which would allow to take advantage of various event triggering and such (`Model::beforeSave()`)
- [ ] Fixtures are self contained. This is an issue as it makes it impossible to add data to an existing graph. Say your user builds a basic graph that he loads via `Migration` he will not be able to add data to that graph and link it to the proper vertices.
- [x] Fixtures should be unloaded/loaded before each test to ensure the state of the DB when you're writing without having to worry about cleanup and such (on failure your cleanup might not even happen).
- [x] Fixtures should be independent from the test suit. If you wanted you should be able to load a fixture for whatever reason. I think `Fixture` should have : 
  - `Fixture::load($db)`, loads the data (duh).
  - `Fixture::unload($db)`, unloads the data (duh)
  - `Fixture::$dependencies`, array of other fixtures this fixture depends on.
  - `Fixture::$data`, path to the data file containing the fixture information.
  
  It's the `TestCase` that should load unload these before each test. Perhaps Tests classes could contain a `Test::fixtures()` method that returns an array of fixtures to load.
- [ ] The fixture data array (`Graph`) should probably follow the `tree` format to be consistent. This would allow us the easily handle the information via spider specific helpers, and even allow users to export a tree query into a fixture data file.
- [ ] Should fixture data be a class? Maybe just a php file along the lines of : 
  
  ``` php
  <?php
  return [/*data*/];
  ```
  
  This would definitely make it easy to export a tree response to a file.
- [x] Perhaps make a separate config file with the configuration array that the bootstrap.php file can use to setup the DBs. That way the Fixtures can be applied to whichever DB configuration is desired. 

---

The way I saw things the dev cycle would have been : 
- Driver
- Builder (with processors)
- Query
- Fixture (requires Query and processors for each db)
- Schema
- Migration
- and then the rest. (Model, Helpers, etc)

Let me know how you feel about everything. I think I took something that was quite simple and turned it inside out. Not sure how deep down the rabbit hole we want to go at this point in time.

electricjones:
Wow, I hadn't realized what a pandora's box I opened, lol. I never intended this to be a final solution, just a temporary normalization. Truthfully, I started out last night wanting to normalize Orient with the other two, off load those expected values, and simplify the DriverTests. One thing led to another...

I have always used the built in fixtures with phpunit and didn't realize they relied on builders.

So, humor me here. A `Fixture` is an array of data, not an actual database? I had thought that a `Fixture` was a test database. In phpunit, I have (in the past) used a test sqlite database to ensure that everything was up and running.

I would have thought a data array that replaces the database for testing would be a mock or stub? And we need to have at least one unit test suite (drivers) that actually communicate with the database, and it would be nice for integration tests to communicate with the database as well.

In any case, I'm happy to scrap the fixture idea, but how can we save the offloading of the expected values to a representation of the graph that all three drivers share? Not only do the driver tests use it, but the integration tests, and ultimately functional tests.

I also have driver credentials to the test databases stored in `Graph` just for lack of a better place. Again, I'd like to be able to update these things in one place and all other tests refer to them. Suggestions?

dmill-bz:
Well the way I see it. Fixtures are a fixed state that ensure a well known and consistent environment. Whether it be databases, files in a folder, preparing data for objects, etc..

So for the purpose of spider lets think of it as the following classes `Fixture`, `DBFixture extends Fixture`, `FileFixture extends Fixture`.
In our case, `DBFixture` should take some input file (an array) + a db and apply what's inside this file to the database (possibly loading other dependent fixtures before hand if required). 

So it doesn't replace the database, it just populates it. The best way of going about this would be by having the fixture generate a `Query` from the data file and submitting that query to which ever driver is configured. This is why I considered it to be a higher level feature. 

So yeah definitely different to a stub or mock.

Again I don't know that we want to put everything in my previous post in right away. But maybe we can do a minor refactoring and then do the rest at a later date. This way it simplifies our testing. 
Seeing as fixtures should be independent from the test suit it should be possible to get this up and running and then refactor it without affecting tests. 

I've checked what is probably the minimum requirement at this stage (in the previous post). If we can completely separate the fixture logic from the test one (just calling load/unload on setup/teardown) then it's enough to go forward with the PR

---

I also wanted to add that we may have conflicts with the slight refactoring I've been doing to dispatch and drivers. (to accept queries and choose the language processor)

electricjones:
Okay, I'm tracking. I definitely don't want to get into true fixtures yet. At the moment, the QueryBuilder won't even create relationships, so it couldn't populate a useful graph.

So, as far as populating fixtures goes, I propose the following as a bandaid for right now:
- `DBFixture extends Fixture` which has a contract for `load()`, `unload()` and setters/getters for data and dependencies.
- `OrientFixture`, `NeoFixture`, and `GremlinFixture` extend `DBFixure`
- Before any test suite that uses the database, the `setup()` method instantiates the appropriate fixture and `load()`s. ATM, this means the `DriverTest`s and the `Integration\WithX` tests.
  - Instead of loading the fixture data from an array file and producing a `Query` (which isn't supported right now), the Fixtures load a vendor-specific script (like the cypher one I used to install Neo4j), cleans the graph, and repopulates it with the new consistent graph.
  - Each fixture-script creates the same graph.
  - On `teardown()` the graph is destroyed.

What I would like to keep (even if refactored) is some version of the `Graph` class that exists now. It provides an easy way to get **expected** data and driver credentials from a single place. This allows the test suites to be much simpler. If you look at the `Integration\QueryBuilder\BaseTestSuite`, you can see how I used that Graph data to find the expected values to test against. And it worked like a charm. Driver tests have also been refactored to use `Graph` as expected data.

electricjones:
I had read an earlier comment wrong. We were on the same pages about what a fixture was, I just hadn't considered that they should be setup/toredown after each test.

But, after a quick google search, it doesn't seem there is an easy way to wipe a database clean in either Orient or Neo4j. In Orient, it's possible to delete the entire db through a command, but there is no Cypher equivalent. How would we wipe neo and gremlin clean between tests?

electricjones:
I think this latest commit is closer to a good solution. That's what I get for halfheartedly doing something, lol.

Anyway, this merges the integration tests (so everything is in one place) and

**Creates a basic `Fixture` convention**
- `Fixtures\Fixture` is abstract with methods to load, unload, get data, and manage dependencies.
- `DbFixture extends Fixture` but adds nothing right now.
- Each driver will have its own fixture class `OrientFixture` which is responsible for a few things:
  - `load($data)` creates the fixture and sets `Fixture::$data` to the expected data. **for now** this is done with a bandaid solution. An orient-specific script is run to build up the fixture. 
  - `static::$data` is set from `Graph::$data` which is just an array representation of the standard test graph. This is used for the expecteds in phpunit. In the future, this data will be variable, loaded from a default file or provided when called. The data will be translated into a `Query` and executed, then set as `Fixture::$data` providing the same access for the expecteds.
  - `unload()` tears down the created test graph. In orient, this means deleting the database. I'm not sure how to pull this off in Neo or Gremlin.
  - `getData()` returns the array representation of the test database for all those expecteds.

**Updates and simplifies Tests**

`OrientDriverTest` and `Integrations\QueryBuilder\WithOrientTest` have been updated and work beautifully. The tests take a while, but whatcha gonna do?

Implementing Neo and Gremlin should be trivial. We just need to implement the `load()` and `unload()` methods. The Cypher script already exists.

This is not a perfect solution. It is a foundation to provide us with better testing now until the Query Builder is up to spec and (maybe) we have some basic Schema functionality.

dmill-bz:
Neo and co have a similar approach to deleting the db. Essentially delete all vertices and edges and you're good. In the case of gremlin-server I know that edges get deleted if one of the connecting vertices is deleted. So in gremlin this should do : 

``` groovy
g.V().drop().iterate(); // drop all vertices.
```

I'm not sure this also works for Neo so to be safe lets remove both:

``` cypher
MATCH (n)
OPTIONAL MATCH (n)-[r]-()
DELETE n,r
```

This approach isn't optimal for huge sets but it shouldn't be a problem for fixtures. 

And I'm cool with the PR appart from the Graph object holding the server info which I think is going to cause problems down the road. It for one doesn't allow us to configure different testing environments (like changing server ports etc..) I mean we can do that but then we end up making changes to a VCed file which is not really desired. That and it goes against the separation of logic you'd expect from a fixture, you should be able to load any fixture in any DB you want. Think of:

``` php
$connection = new Connection([
   'db1' => [...],
   'db2' => [...],
   'db3' => [...],
]);
$fixture = new Fixture();
$fixture->load($connection->db1);
$fixture->load($connection->db2);
$fixture->unload($connection->db3);
//or
$newFixture = NewFixture(['db'=>$connection->db1]);
$newFixture->load();
$newFixture->unload();
```

It doesn't make much sense for the fixture to have any DB information.
Ideally we would use the connection manager to configure the various DBs for the test suite and then pick out of that hat. 

I think the idea here is that fixtures can be very rudimentary at this point that's not a problem. But it should be sufficiently abstracted that we can change anything we want (appart from the Fixture interface) without having to change the tests. And in this case having the Graph object called directly from the tests is an issue. There's also the question of `getData()` returning consistent data as well. 

I'm not sure whether to take care of this now or later to be honest. Right now it's only used with orient so it's fine. But once we start implementing fixtures elsewhere it would make sense to get it sufficiently abstracted.

electricjones:
I don't mind moving the server information out of the `Graph` object. I just didn't want to copy and paste that information between tests, which seems to go against the DRY principle.

What do you mean by `getData()` returning consistent data? In the future, `getData()` will return whatever data was used to setup the fixture and there is no `setData()`, so there is no way to modify it. For the moment, `getData()` returns an array clone of the fixture, but still no way to change it.

dmill-bz:
Well what I mean by that is that if we decide to have fixture data in `tree` format it shouldn't break the tests. Maybe this means tests shouldn't use getData at all but I guess the other option is to decide on the format early on and not change it.

electricjones:
If the tests are using `getData()`, then how do they get the values for all the expecteds? Especially if it is possible to load different fixtures.

This is probably a problem for down the road.

dmill-bz:
Well I think expected results should be hard coded instead of referring to the fixture. The modern graph fixture should be complete enough for 99% of our test scenarios. And for the others we can always make a fixture of a more complex graph. In any event we will never have to alter the modern graph for future tests so it's safe to assume it will never change.

electricjones:
Alright. I don't know why that feels so odd to me. But, if its best practice, I can just chalk it up to learning something new. I just rebased. I am going to remove the `Graph` and hardcode the credentials and expected results throughout the system.

For now, each `DriverFixture` will have a hard-coded script to setup and teardown the database fixture. Once we get a QueryBuilder on its feet, we can refactor to use a config file with array data and the QueryBuilder.

dmill-bz:
That sounds good! 
At one point we will have to decide how spider places in terms of library vs framework. ie: will we ever have a spider entry script? Or will we ever provide some means of setting up local non VCed configuration files. 
The correct answer is probably to make the above optional, not sure what the best way to go about this is though.

electricjones:
Can you provide me with a gremlin script to populate modern-graph?

dmill-bz:
There are two elements to graphs in gremlin server (all in the configuration). The graph object (often `graph`) and the traversal object (often `g`).
In the case of our DB setup there are two graphs. One non-transactional `graph` with `g` (that auto loads with the modern graph) and one that supports transactions `graphT` with `t` (neo4j that loads empty). Maybe we will need to readjust these for the future.
Anyways the code is:

``` groovy
TinkerFactory.generateModern(graph)
```

Cleaning this up would be 

``` groovy
g.V().drop().iterate()
```

electricjones:
I have implemented the above and created the fixtures for both Gremlin and Neo4j.

There are a few issues to be resolved:
- [ ] The ids of the fixtures are auto-incremented, which means the that checking the ids in the tests don't work. For now, I have just commented out the assertions that test against IDs. I don't think we need them.
- [ ] For some reason, the Neo4jDriverTest::testFormatPath() is failing because the results from the fixture and they don't match with the expecteds. Can you take a look at that?
- [ ] I have also followed the same process on GremlinServer, but still don't have it running on my system. A quick check to make sure those tests pass would be appreciated.

Now, the way everything works:
- `Fixture` contracts `load()` and `unload()`
- `DbFixture extends Fixture` but adds nothing yet
- `<Driver>Fixture` implements load by creating a client object, populating with modern graph (hardcoded for now). They implement `unload()` which wipes modern graph clean
- Query builder integration tests have begun. `WithOrientTest` is active. Once processors come along for the others, it will be trivial to add the to the list
- `DriverTest`s utilize fixtures, save the bugs above.

dmill-bz:
ok I'll give this a spin

electricjones:
I have opened up issue #49 to discuss a roadmap for fixtures of the future. What we have works for now, but we can update fixtures as spider evolves.

dmill-bz:
FYI the neo issue has to do with the ordering of ids. Neo seems to just randomly attribute IDs which messes with the order. I'll see what I can come up with to fix this test.

electricjones:
If it is easier, we can just fabricate a response that can be formatted as a path instead of doing an actual query.

dmill-bz:
ok I just pushed gremlin and neo correction. I also changed the loading/unloading behavior of fixtures. Now setup unloads first (incase there's anything left over from a crash/fatal error)

dmill-bz:
I think this is good to go

PR #43, opened 2015-08-13
Adds integration tests for QueryBuilder

electricjones:
See #39, #38 

This PR lays the foundation for integration testing. When I say integration tests, I mean tests that test the functionality of several components together, but does not necessarily look into the database, and may be more specific that a functional or acceptance test. I realize terminology can be different from different sources, but that is what I have decided to stick to.

There is a place for functional testing (which can peer into the database) and perhaps acceptance testing (which is only aware of responses, no internals).

In any case, these integration tests:
- Extend a BaseTestSuite which holds all the actual testing logic.
- All an implementor is required to do is setup `$this->query` as a querybuilder with a valid connection. See `WithOrientDriver` as an example.
- Since Orient is the only one with a LanguageProcessor, it only tests Orient. It will be trivial to add the others once processors are complete

These tests are incomplete. They cover some basic select, insert, and delete. No update. At the bottom of the tests are a list of methods that need to be tested.

Note that _all of these methods should already pass_, we just need some tests written. Basically copy and paste the above tests and modify, slightly. I am asking for help on this, but will not wait to merge. We can (and should) add new tests as we go along.

This breaks nothing. It does include two very minor changes I ran into through the process:
1. Adds an alias of `Query:;dispatch` called `Query::go()`. Just makes it easier to do `$query->insert()->data()->into()->target()->go();` and the like
2. Fixes a bug that was keeping the `Connections\Manager` from creating connections based on an alias driver name. Adds tests for this patch.

**Last Note**: This is branched from the `orient-fixture` branch and requires those commits. However, if orient-fixture is merged before this one is ready, I will simply re-issue the PR against develop. There are no merge conflicts. This only adds some new stuff.


dmill-bz:
looks good. I have some comments but I'll put them in the main PR

Issue #42, opened 2015-08-13
Inserting multiple records via the QueryBuilder


electricjones:
As of now, it is only possible to insert a single record per query in the query builder.

``` php
$query->into('target')->insert($data);
// or
$query->insert()->data($data)->into('target')->go();
```

It should be easy to insert multiple records at once.

``` php
$query->into('target')->insert([
   [ 'name' => 'michael']
   [ 'name' => 'jonothan']
]);
```

Something to keep in mind, though, is that some databases allow nested properties, like OrientDB which is also a document database. Therefore, the processor should check to see if the `$data` in the `Bag` is an array of arrays AND that the `$data` array contains only indexes (0, 1, 2, etc) and not key-value pairs.

dmill-bz:
Yeah there's a lot of stuff to think about here given property cardinality and such. Both of these should be possible and be different:

``` php
[
   'name' => 'elizabeth',
   'name' => 'liz' // although this isn't valid PHP it's valid as far as data storage goes for some graph DBs.
]
```

and

``` php
[
   'name' => ['elizabeth', 'liz']
]
```

electricjones:
The issue with the second one is in the Language Processor deciding if we want two records, or one record with two names (which is supported in Orient, at least).

electricjones:
I would probably suggest restricting multiple inserts to:

``` php
[
     ['name' => 'elizabeth']
     ['name' => 'liz']
]
```

But we can add sugar to the QueryBuilder to make it more palatable. I just feel like a language processor would know exactly what to do with the above.

dmill-bz:
Well in my two examples the first one was meant to be one record with two String names, and the second example was one record with one array name (containing two names). Both of these are possible.

There's a third case as well. That the record would have two "name"s that are both an array of several other items.
So it becomes quite complicated.

dmill-bz:
here's what I'm trying to get to:

Case of multiple inserts (multiple vertices for example):

``` php
[
   ["name" => "something"],
   ["name" => "other"],
]
```

Case of the name property having a cardinality of `single`:

``` php
[
   ["name"=> "something"],
]
```

or even though the cardinality is `single` the value can be an array:

``` php
[
   ["name"=> ["something", "other"]],
]
```

Then the case of a cardinality of `multiple`

``` php
[
   [
      "name"=> "something",
      "name"=> "other"
   ],
]
```

or the same with arrays:

``` php
[
   [
      "name"=> ["something", "other"],
[...]

PR #44, opened 2015-08-14
Multiple inserts per #42. does not handle cardinality

electricjones:
See issue #42 

This now allows for (in the orient processor):

``` php
$builder->insert([
                [
                    'name' => 'mal',
                    'role' => 'captain',
                    'ship' => 'firefly'
                ],
                [
                    'name' => 'zoe',
                    'role' => 'first',
                    'husband' => 'wash'
                ],
                [
                    'name' => 'book',
                    'role' => 'shepherd',
                    'past' => 'unknown'
                ]
            ])->into('target);
```

It does not handle cardinality, but I like the DTO idea, generally.


dmill-bz:
This looks good. I'll start working on a cypher processor in the next few days so I might have more feedback by then, but for the time being it looks safe to merge.

PR #45, opened 2015-08-14
made multiple changes to how queries and drivers interract

dmill-bz:
There are a few things covered in this PR:
- [ ] `Builder::getScript($processor)` now returns a string
- [ ] `Builder::getCommand($processor)` replaces the old `getScript()` (I've also moved both up to the `BaseBuilder`.)
- [ ] I removed `getRw()` and `setRw()` as these weren't necessary. The `Command` doesn't have any authority  over the driver. It's now the `Query` or user's job to decide which scenario (R/W) we're in and act accordingly.
- [ ] `Query` no longer requires a `Processor`. It now delegates the processing work to the Driver. If you still decide to provide a processor to the query then that one will be used instead and a `Command` will be passed to the driver.
- [ ] a set of methods in `AbstractDriver` help the drivers convert `Builder` into `Command` depending on available languages. 

---

Things to keep in mind:
- The current implementation of converting `Builder` to `Command` in the drivers is verbose and an ugly copy/paste. We can change this once we have the other processors up and running, we'll just create a method in `AbstractDriver` to handle this. Should make the code much cleaner. 
- Other PRs will need to rebase around this one. (I think it'll be easier this way around rather than rebasing this against the other PRs)
- Currently have a very simple Builder to Command Driver test for Orient. We will need to move this up to `BaseDriverTest` once we have at least a cypher processor.
- There are no added tests here for `Query`. Main reason being that the best place for this would be in the integration tests and that's in another PR. We can add it after as we go through code coverage.

### DIFF
--- a/src/Commands/BaseBuilder.php
+++ b/src/Commands/BaseBuilder.php
@@ -2,6 +2,7 @@
 namespace Spider\Commands;
 
 use InvalidArgumentException;
+use Spider\Commands\Languages\ProcessorInterface;
 
 /**
  * Fluent Command Builder, simple bag manipulation
@@ -282,7 +283,7 @@ class BaseBuilder
      * Return the current Command Bag
      * @return Bag
      */
-    public function getCommandBag()
+    public function getBag()
     {
         return $this->bag;
     }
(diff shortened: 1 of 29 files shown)


chrismichaels84:
So, if `Command` no longer has a read/write property, how does the driver know which to execute? Is it up the the driver to do a search the script for keywords? That could be unsafe.

Issue #54, opened 2015-08-17
Improve Test Coverage


electricjones:
Right now, our test coverage is suffering from a lot of little things and a few big things. I'd like to tackle a few before we get too far ahead of ourselves.

electricjones:
PR #55 cleans up the test coverage a lot, but there are some bigger concerns that I just didn't want to deal with right now. Would love some outside help. These should be easy:
- [x] Remove `Graph\ID` entirely. It's only referenced in a test or two to normalize the way record IDs are handled by the different database. There has to be a more elegant way to do that.
- [x] Improve testing around the `formatAsX` methods in the Drivers. Granted, this will come once we have settled on formats
- [x] test `runRead` and `runWrite` commands

dmill-bz:
I've partially removed `Graph\ID` already in my upcoming PR. I'll elaborate once I push everything.

dmill-bz:
PR #56 covers the `Graph\ID` case.

Thought of something but the `runRead` makes little sense since you want to get something back from the DB when you're reading. Maybe we can remove this altogether, and instead of `runWrite` we should just call it `run` (the write nature being implied).

electricjones:
Totally agree. runRead serves no purpose. I like run. I'll take a look through the rest of this tomorrow.

electricjones:
I'm closing this. It'll come as we move forward.

PR #55, opened 2015-08-17
Small changes to improve test coverage

electricjones:
See PR #54 

This is a minor PR that attempts to up the test coverage without making any significant changes. I just wasn't happy with the coverage report, and it showed me a few legacy artifacts that were still hanging around.

For the most part, this is all super minor:
- [x] Removed `Spider\Graphs` for now. We can revisit them when we need to. I did keep `ID` for now because its used in a test, but it should probably be removed as well. I'm just lazy right now.
- [x] Removed a good chunk of `QueryTest` in unit tests. I'd bet all of this should be moved to the Query Integration test, but I'd have to look more closely.
- [x] Added incedental tests for setters, getters, etc.
- [x] Added an AbstractDriverTest that tests `isSupportedLanguage()` and `getProcessor()`
- [x] General cleanup of some docblocs and comments in tests

There were a couple very minor tweaks to the code:
- [x] `AbstractDriver::getProcessor()` throws exception if unsupported, with test
- [x] Removed `make()` from `CommandBag`. It served no purpose
- [x] Removed `DriverInterface` from `Connection` and `ConnectionInterface`. I'm torn about this one and can put it back if it matters. All the calls are already passed down to the driver magically. The only reason `ConnectionInterface` even extended `DriverInterface` was for auto-generated API documentation. That is still important to me. If we do keep it, we will have to write some basic automated tests. It was a pretty big hit to the overall coverage score.


dmill-bz:
I really like this PR. I think Connection should follow a different logic to that of the drivers, I quite like the change actually. I think it'll make more sense once we implement transaction emulation logic, caching, and some other features.

Issue #50, opened 2015-08-16
Cypher processor


dmill-bz:
Hey I've got questions in regards to this. I figured we could open an issue while I finish this dev.

I'm currently just going through the tests to make a failing Cypher processor test and I wanted to know about this query:

``` php
$query = 'SELECT field1, field2';
$query .= ' FROM target';
$query .= $this->getWhereSql();
$query .= ' GROUP BY groupField';
$query .= ' ORDER BY orderField DESC';
$query .= ' LIMIT 3';
```

The logic seems a little off in the sense that you are grouping by groupField but ordering by another field entirely (that would be lost in the grouping process). It's important to get the logic right here as cypher doesn't have a `GROUP BY` clause which means that I need to construct an equivalent.
Same with the selecting of fields that would be lost in the grouping process. We should change this around to be something more realistic. Maybe just drop the `GROUP BY` and place that in another test?

electricjones:
You are right. I was just trying to create a complex query bag and wasn't really thinking about a practical scenario. I am fixing the BaseTest and Orient Test now. Should I push to develop or would you rather a pull request (don't want to screw with what you are working on now)

dmill-bz:
The other issue is that there should be a clear separation of logic between label and type.

In both gremlin and cypher you choose the type of element you want and then the label within that type. 
ie: I want a `vertex` with label `target` and/or an `edge` with label `target`, or both. The query for these would be different. (whereas I feel it would be the same query in orient).

This means that the use of `from()` is ambiguous at best. There's no intuitive understanding of what it refers to and technically in graph DBs it doesn't have much meaning other than referring to a subset of the graph (which is what `where` is for). I understand that in orient they've made label/class analogous of RDBMS tables because it made the transition easier, but this is an incomplete abstraction and doesn't represent graph logic properly. 

At best I could see from() being used as an optional field to specify if we're looking to get `vertices` or `edges` (as opposed to the default `both`)

There could be several ways of going about it. Here are a few:

``` php
// using where. This could work for IDs as well (as we discussed ID cases in one of our previous exchanges)
$builder->select()->where(Builder::ELEMENT_LABEL, 'user'); // fetches user labelled elements (both V and E) 
$builder->select()->from(Builder::ELEMENT_VERTEX)->where(Builder::ELEMENT_LABEL, 'user'); // fetches user labelled vertices 
[...]

PR #56, opened 2015-08-18
Added a cypher processor and made some structural changes to builder

dmill-bz:
Ok so this ended up being heavier than I thought. Let me go ahead and give you the bad news first:
- [x] I completely broke the orient processor with my changes. Figured you could have a look and correct things faster than me here.
- [ ] We have a bit of a (big) structural issue in query building. Neo and Gremlin allow (and for some actions require) a combination of `select`, `update`, `create`. Which is not something we can currently do.
- [x] Right now `BaseBuilder` has some level of sugar syntax (`where()` for example) we should move this down to `Builder`. The association of Bag::X and string versions should probably go down as well. I've done some of this but not all.
- [x] I've made some changes in this direction but I think there are more: `Builder` sugar should allow users to simplify how they pass the data to the methods, not allow them to add different types of information. An example here is update() that could take both `data` and a `target`, this really complicates the API and code. For a more intuitive use we should only allow one "type" of information in different sugared formats. (Don't know if I'm making sense)

---

Now here's what this PR has:
- [x] I removed `BaseBuilder::target()` as it was ambiguous. In the same line of thought `Bag::$target` now refers to the element type (Vertex via `Bag::ELEMENT_VERTEX` or Edge via `Bag::ELEMENT_EDGE`). I kept `target` as the name but we can change this.
- [x] Introduced a `Builder::type()` method to define the type.
- [x] labels are now to be set in `Bag::$where` and `Bag::$data`  with a key of `Bag::ELEMENT_LABEL`. The processors will have to translate this where row to the appropriate language spec (I think only orientSQL will need special care here).
- [x] Introduced `Builder::label()` to automate this.
- [x] IDs are now to be set in`Bag::$where` and `Bag::$data` with a key of `Bag::ELEMENT_ID`. The processors will have to translate this where row to the appropriate language spec (I think only orientSQL will need special care here).
- [x] Changed `Builder::record()` to reflect this.
- [x] Changed how ordering worked. We had only one ordering direction for all ordered fields. Now you set order values via `Builder::orderBy('field', Bag::ORDER_ASC)` (default), and the `Bag` stores this information as `[['field1', Bag::ORDER_DESC],['field2', Bag::ORDER_ASC]]`. I had to remove some of the sugar for this, we can discuss how to sugar this new signature. 
  
  I removed `Bag::$orderAsc` in the process as well as the `Builder::asc()` and `Builder::desc()` methods.
- [x] Changed `BaseBuilder::update()` to only accept data. Target info should be added via `record`, `from`, `label`, `into`, etc.
- [x] Added `Bag::COMPRATOR_IN` for array scenarios.
- [x] Added `Bag` conjunctions for `XOR` and `NOT`scenarios.
- [x] removed the `src/Graphs` folder and references to the no longer needed `ID` class.
- [x] made `Bag::$data` consistent. It should now always be a double array regardless of the number of elements `[['data'=>'here']]` or `[['data'=>'here'],['more'=>'here']]`. It's the Builder's job to ensure the bag has the correct format now, not the processor's job to figure out what the format is.
- [x] Renamed `BaseBuilder::create()` to `BaseBuilder::insert()` to simplify the API. This way people checking the `Builder`API don't have both methods show up with different signatures (less confusing).
- [x] Added a couple of processor tests around handling edges instead of vertices. This is way incomplete but we can make this better in one of the upcoming code coverage PRs we'll be doing.
- [ ] Changed some parts of the documentation to reflect some of these changes

Let me know if you see anything outrageous!


electricjones:
I am happy to fix orient tests. I know orient pretty well, so that shouldn't be a problem. May I assume that all the Neo tests work as it is?

dmill-bz:
Yes they all work. Only orient processor and integration tests fail (and one driver test that should auto fix once the processor is up to date)

electricjones:
Alright, this all looks good, but I do have a few questions/counter-proposals/challenges/discussion-thingies. I will pull this tomorrow and try to get the orient processor up to snuff and really dig into each of the above, then respond a chunk at a time.

I really like the direction this is going. I do have to say that I agree the `BaseBuilder` should be as straight forward and rigid as possible, but it is important to me that `Builder` be able to be used fluently. Code that tells a story is just easier to manager.

dmill-bz:
As a reminder we can close #50 once we've agreed here.

electricjones:
Alright, I've been fighting with travis and coveralls, and finishing up another deadline. The rest of my week is (more-less) freed up for this. I'd love to get to 0.3 or really close sometime Sunday, as I leave town for a week on Sunday night.

That said, here are some of my initial impressions about this PR.
- I've already checked off the items I'm totally onboard with.
- Fixing the orient tests should be no problem
- This moves `getCommand()` and `getScript()` to the `BaseBuilder`, but `BaseBuilder` has no inherent awareness of a processor. We had originally said the `BaseBuilder` does NO processing, only manipulates the `Bag`. I don't mind `BaseBuilder` processing, but that should be reflected throughout, i.e. in the constructor, getters, setters, fields, etc. If you agree, I can do this easily. If not, what justification was there for this?
- I don't understand why `where()` should be moved to `Builder`. That, to me, isn't sugar, its necessary. `BaseBuilder` isn't abstract, we should be able to use `BaseBuilder` for the entire process, even if it is verbose and unable to process into Commands or Scripts.
- Can you give me an example of when `select`, `update`, and `create` would be used in a single operation? Are you talking about sub-queries? 

``` php
$builder
   ->update($data)
   ->where(new Builder()->select()->from()->where('user', 'username'));
```

In either case, Orient allows for SQL Batching, which is how I handle transactions, anyway. I don't think this will be a problem.

I'm sure I'll have more questions as I dig into the tests. For now, I'm making changes in a local branch. It'll probably be best if I issue a PR against this branch instead of pushing directly.

On some side notes: I've run into more complications with gremlin-server but I'm hunting down stack traces now. This time its while installing the hadoop dependencies.

I've also setup Travis and executed the first build, which actually failed for some of the jobs but not others. You can see at https://travis-ci.org/spider/spider/branches.

electricjones:
I have issued a pr to bring orient into this pr. I didn't touch anything except orient and the integration tests, with the exception of updating a few incompatible signatures.

This is more-less ready, but I'd still like to discuss the items I mentioned.

dmill-bz:
> This moves getCommand() and getScript() to the BaseBuilder, but BaseBuilder has no inherent awareness of a processor. We had originally said the BaseBuilder does NO processing, only manipulates the Bag. I don't mind BaseBuilder processing, but that should be reflected throughout, i.e. in the constructor, getters, setters, fields, etc. If you agree, I can do this easily. If not, what justification was there for this?

Yes I agree. I'm not sure what thought process led to not including the processors initially but they should be part of baseBuilder.

> I don't understand why where() should be moved to Builder. That, to me, isn't sugar, its necessary. BaseBuilder isn't abstract, we should be able to use BaseBuilder for the entire process, even if it is verbose and unable to process into Commands or Scripts.

I think `where()` should be in `BaseBuilder` in it's initial version (no handling of sugar). We can then override it in `Builder` to add support for sugar syntax (and convert it before sending it to `parent::where()`)

> Can you give me an example of when select, update, and create would be used in a single operation? Are you talking about sub-queries?

No (or at least not on the surface, maybe it can be fixed with subqueries). One such example would be selecting two nodes to create an edge between them. But you can imagine way more complex scripts that mutate the graph during traversal (many neural network user cases would need this).

> On some side notes: I've run into more complications with gremlin-server but I'm hunting down stack traces now. This time its while installing the hadoop dependencies.

Let me know if you need any help. It will probably save you quite some pain.

> I've also setup Travis and executed the first build, which actually failed for some of the jobs but not others.

I pushed a correction for this, it now passes (I didn't give neo4j enough time to init before trying to change the password)

electricjones:
PS I am almost done with a PR that moves all `Query` tests to integration, covers the full api (I think), and tests Neo as well. I should have it ready tomorrow morning. After that, this will probably be ready to merge.

electricjones:
Alright, I track with everything above and agree totally. Not sure what to do about single operation CUD queries. Orient has batch SQL, so that should work pretty well on that end. What did you have in mind?

dmill-bz:
I'm not really sure at this point. I was thinking of doing something along the lines of taking a language grammar.. say the following cypher grammar:

``` cypher
MATCH ...
OPTIONAL MATCH ...
WHERE ...
ORDER BY ...
LIMIT ...
CREATE ...
SET ...
RETURN ...
```

Break it down into common sections like say:
- `SELECT` = `MATCH` & `OPTIONAL MATCH`
- `WHERE` = `WHERE`
- `ORDER` = `ORDER BY`
- `LIMIT` = `LIMIT`
- `GROUP BY` = Something the processor would have to translate since there is no cypher equivalent
- `INSERT` = `CREATE`
- `UPDATE` = `SET`
- `PROJECTIONS` = `RETURN`

Store them separately and independently of each other and then piece them together according to the grammar layout. 

In that sense we would have : 
- `SELECT` that contains type information (V or E)
- ...  (skipping the self explanatory ones)
- `INSERT` would contain all information about what elements we're creating (V or E), how many of each and that's it (depending on how you look at it an insert could be a creation of element and then an update of said element.
  
  This would be transparent to the user (insert would still work like it currently does) we would just store it differently. `INSERT` would have the info mentioned above and somehow linked to the corresponding data in `UPDATE`
- `UPDATE` would contain the current `$data`. That is, fields and their new values (whether it be for creation or update)
- `PROJECTIONS` is the same as currently. It can be populated as it currently works.

The thing is. This can become very tedious. Maybe it's best that we decide on how to write a simple query that selects two vertices and places an edge between them. Then go from there.

Here's one suggestion: 

``` php
$builder->select('c')->where(['a.name'=>'marko', 'b.name'=>'chris']) // builder understands we're getting two sets a and b
        ->limit(2)
        ->insert(['c.type' => Bag::ELEMENT_EDGE]) // new c element
        ->update('c')->withData([Bag::ELEMENT_OUTV=>'a', Bag::ELEMENT_INV=>'b', Bag::ELEMENT_LABEL=>'friend'])
```

I don't like this for many reasons but man am I at a loss trying to figure something better out. Well hopefully this gets the ball rolling. We should probably open an issue for this.

dmill-bz:
The other thing I can think of is that these mutating traversal scenarios often depend heavily on traversal logic and should probably require using a `TraversalBuilder`. In which case we could "say screw this" on the Builder layer, and keep it simple. the above and some other simple scenarios could be accomplished via:

``` php
$builder->insert()->type(Bag::ELEMENT_EDGE)
        ->withData([
            Bag::ELEMENT_OUTV=>(new Builder)->select()->where(['name'=>'marko']) ,
            Bag::ELEMENT_INV=>(new Builder)->select()->where(['name'=>'chris']),
            Bag::ELEMENT_LABEL=>'friend'
        ])
```

This keeps the Builder API simple but doesn't change the fact that we will need to change the `Bag` structure.

More complex scenarios can be run using the upcoming `TraversalBuilder` and will be much tidier due to the imperative nature of that builder.

electricjones:
A few points here in response and kinda related.

---

So, are you suggesting a `Bag` layout along the lines of:

``` php
class Bag {
     $command; // same as now
     $insert = [
            Bag::ELEMENT_OUTV=>$builderInstanceToBeProcessed, // Using a sub-select
            Bag::ELEMENT_INV=>$AnotherBuilderInstanceToBeProcessed, // Using a sub-select
            Bag::ELEMENT_LABEL=>'friend'
     ],
     $data = [[]]
}
```

Because I like that separation.

I also like using dot notation whenever we can. It just makes sense.

---

I had figured we could (in our next iteration) embed instances of `Builder` inside `Bag` and then `Processor` could do a simple type check and process into a `Command` to be inserted into the parent command.

Also, we should introduce a simple `insertEdge()` at the `BaseBuilder` level to make explicit when we are creating relationships. All other RUD operations can use the standard.

``` php
// insertEdge($label, $from, $to, $properties = [], $direction = '')

$baseBuilder->insertEdge(
    'is_following',
    new Builder()->select->from('users')->where(['username', Bag::EQ, 'chris'])->limit(1),
    new Builder()->select->from('users')->where(['favorite_color', Bag::EQ, 'purple'),
    [
        'date_followed' => 8/20/2015,
        'weight' => .5
    ],
    Bag::DIRECTION_OUT
);
```

Now, chris is follow everyone whose favorite color is purple.
Which could obviously be sweetened at the `Builder` level and dispatched at the `Query`.

---

I was thinking about the **Traversal Builder**, and I think our best option is actually to follow Orient's lead in this case. They added a `TRAVERSE` command to standard SQL:

`TRAVERSE * FROM #10:1234`
`SELECT FROM ( TRAVERSE friends FROM #10:1234 WHILE $depth <= 3 ) WHERE $depth >= 1`
`SELECT $path FROM ( TRAVERSE out FROM V WHILE $depth <= 10 )`

In many cases, though, SELECT works much faster than TRAVERSE:
`SELECT out('follow').out('follow') FROM TwitterAccounts`

Though, in our case, I don't know if I like mixing the imperative/declarative aspects. Tread carefully.

For spider, what if we simply included the Traversal Builder as a _dependency_ of the `Builder` or `Query`. This allows the user to not worry which builder to instantiate or configuring two builders. The user simply: `$query->traverse()->match()->out()->out()->go()`.

I like the idea of using Cyphers terms for traversals, they seem clearer. But that may confuse the issue.

That would also allow us to inject the standard builder into the Traversal Builder and perform subselects in a similar way that Orient does.

_Get all the friends up to the third level of depth_.

In OrientSQL: 
`TRAVERSE friends FROM (SELECT FROM person WHERE username = 'dylan') WHILE $depth <= 3`

``` php
$query->traverse()->start(
    new Builder()->select()->from('person')->where(['username', Bag::Equal, 'dylan'])
)
    ->while(['depth', Bag::LE, 3])
    ->all();
```

Or with sugar:

``` php
$query->traverse(
            'select', ['from' => 'person', 'where' => ['username', 'dylan']]
     )
    ->while(['depth', '<=', 3])
    ->all();
```

Which could probably be sweetened further.

dmill-bz:
Well now that I've had a good night's sleep and thought about it some more. We might be able to get away with it without changing Bag. 

If we decide to go with : 

``` php
$builder->insert()->type(Bag::ELEMENT_EDGE)
        ->withData([
            Bag::ELEMENT_OUTV=>(new Builder)->select()->where(['name'=>'marko']) ,
            Bag::ELEMENT_INV=>(new Builder)->select()->where(['name'=>'chris']),
            Bag::ELEMENT_LABEL=>'friend'
        ])
```

Then we will end up with a `Bag` that contains other `Bag` elements, all of which function like the current `Bag`. I also think that this scenario is better than having `Builder` inside `Bag` (the `Processor` shouldn't know what to do with a `Builder`). And it has the advantage of being a good way of easily implementing nested queries (aka traversals in a declarative way).

With that said, the traversal scenarios that are more complex might require that separation of logic (but maybe we should worry about that when we get there)

---

As far as **Traversals** go (imperative kind) we should probably use gremlin as a basis as.. well that's the reference language for these kinds of operations. 

There's going to be a lot of thinking required here. For instance, your example of while looping doesn't specify what actions to loop. Also the mix of imperative and declarative style commands will be messy. Here's an example:

``` php
$builder->traversal()->start(/* ... declarative Builder ... */)
        ->repeat((new Builder)->traversal()->both())->times(3) // or ->until( <some bool clause> )
```

In this case is `(new Builder)->traversal()->both()` a correctly formatted `Bag`? This will have to be a traversal Builder and not a regular builder.. So we're mixing both types in one query. Going to be tricky.

electricjones:
I really like that insert data. We need to add properties to it, but that's just a fourth key.

I feel like that logic should be in the `BaseBuilder` and then abstracted further in the `Builder`. Somewhere, I would like `$query->createEdge($label, $from, $to, $direction, $properties)` which can be translated into your logic and stored in the bag. We could also try `connect()` and `relate()`, which I've seen in other libraries.

But all of this is for the next iteration, yes?

For imperative traversals, model after gremlin. I was saying for declarative traversals, we can mix in some Cypher as needed. But again, that is the next cycle.

I'd like to focus on what needs to be done to merge my other pr into this and this into develop. Then cleanup and release `0.3`. Prune branches and issues. If we can get that far by the end of the week, when I leave town for a few days, my team can start to integrate this into the app we are developing while I'm gone. Real world testing, even if limited.

dmill-bz:
Yeah those are all for the next iteration imo. This is good to go

PR #58, opened 2015-08-19
Travis dbs

dmill-bz:
This PR covers setting up Neo4J, Orient, and gremlin-server in travis. It also removes the `markedAsSkipped` from driver tests.
I've set up nightly PHP builds to be optional (not failure binding) as it can be a little unstable and in our case one of the neo libraries fail. 
Other than that all tests pass for the other builds so this should be good to go. The suit takes about 5mn to run so it's still acceptable (can't really do better on a free travis account). 

It does **not** cover anything beyond the unit tests. So no coveralls or scrutinizer, etc. We will need to add those hooks in.


smolinari:
Nice! :+1: 

Scott

PR #61, opened 2015-08-20
Sures up integration tests

electricjones:
This pr
- [ ] Fleshes out the integration tests for the query builder. It moves all the `Query` tests from Unit and puts them into a single `QueryBuilder` integration test.
- [ ] Had to work out a couple orient bugs, but all is well now.
- [ ] Created a `WithNeo4jTest` for querybuilder integration testing with neo, but it fails with cypher syntax errors. Don't know if its the builder, the processor, or the test that is the problem. I'm gonna familiarize myself with cypher.
- [x] Now the Neo driver fetches the appropriate Language Processor if handed a `BaseBuilder`
- [x] Removed `updateFirst()` from builder. It's needless, I think. We can always bring it back.
- [x] I also removed `command()` from QueryBuilder. It's purpose was to allow people to issue native queries through the querybuilder, but I don't know how to assign whether it is a write or read query through that method. The `Command` has no awareness, and the `Driver` doesn't know which one to send it through. For now, I just commented it out.

On that note, I will look into SQL batch. It may be a better idea to send everything through as a batch. Since Orient is the odd duck about read/write commands, that may remove the need for the distinction and simplify the driver api. I'll take a look at it in the next cycle.

I've also updated some documentation, but it'll need a good once-over before release.

So you know, here are the dispatch methods and return formats for `Query`:

| Method | Response format |
| --- | --- |
| `dispatch()`, `go()` | `Response` |
| `all()` | Always an array. Limit of `false` |
| `set()`, `get()` | Set (array or single collection) |
| `one()`, `first()` | Set with a limit of one, so a single `Collection` or empty array |
| `scalar()` | Scalar value |
| `tree()` | Tree format (unsupported as of yet) |
| `path()` | Path format (not fully supported as of yet, Orient) |
| `insert(null)` | `Response`. Only dispatches if null is passed |
| `drop(null)` | `Response`. Only dispatches if null is passed |

command($command)       Response (allows you to execute a Command, but you have to build it)
    Does not work -- commented out for now


dmill-bz:
This is looking good. I don't have much to add. If you need help with the Neo end of things I'm more that happy to add to this PR. 
Oh, I do however think that `update()` and `insert()` should require a `go()` or `run()` instead of running even if _null_. I can picture this being a little confusing for people who are used to writing `$builder->insert(/* ...*/)->go()` suddenly getting an error when writing `$builder->insert()->go()`

electricjones:
I can live with that. I'll make that change and them I'm off to lunch. If you can peek at the neo integration tests that would be great. Maybe narrow down where the issue is. I need to familiarize with neo, anyway.

You are welcome to push to this branch. I'll pull appropriately.

electricjones:
Removed dispatching from drop and insert with tests and documentation.

Also, fyi, the integration tests run really slow. I had to setup the fixture after each `specify()` block to ensure proper testing, and this slows things down.

electricjones:
The bulk of the problem seems to be in some minor cypher syntax errors. I'm tackling them now :smile:

electricjones:
Ok, I have squared away most of the issues and squashed a bug or two across the entire bored. Though, in a couple places, I did a quick and dirty job. I don't know if it will cause problems with traversals. All in all, I changed very, very little.
- Cypher likes LIMIT to be after the RETURN for queries
- inside `NeoProcessor::buildProjections()`, I had to reset the variables after the projections were built. They were carrying over to the next query. Not sure if this is the best way to fix it, but it passed tests.
- For `NeoProcessor::buildWheres()` in the case of an element id, the where was setup with a `Bag::IN` comparator, but that didn't exist in the operator map. I hardcoded '=' in that case because I didn't know the correct equivalent (if it exists). Processor line 310
- When CREATEing elements in cypher, the label was left off. I did a quick string replace at line 92, but this may break future cycles. Idk

Last, the update and drop tests for the QueryBuilder integration tests are still failing, and I can't see what's happening. I created an override in `WithNeo4jTest` just to watch everything, but can you take a look at these failures? The same problem applies to the failed test in `insertAndDeleteByIds`

Everything else looks good across the bored :)

dmill-bz:
I reverted your cypher changes as they introduced other issues. I've fixed everything up. A notable change is that I removed `Builder::into()` as it was ambiguous. It was treating labels like tables instead of meta-properties. It also meant that we had to pull the information out of `where` to add it to `data` which was just weird.

Also the drop tests are a little awkward. Normally dropping a vertex means dropping all edges coming/going from/to the vertex, then deleting the vertex. Some DBs do this automatically others don't. The test case should assume the later (as it wouldn't break the former). 
Currently:
- Cypher drop tests provide the correct query to the DB, but the query fails because the vertex is attached to others via edges that weren't deleted beforehand. This is actually the desired behavior.
- The Cypher processor **should not** provide logic for removing edges then the vertex. This should be done in a transaction so belongs on a higher level.
- Even if we changed the integration test to the proper behavior (deleting edges then vertex), this requires traversal logic which we haven't implemented yet. So I skipped this test for neo for now.

I'm good to merge this if you are.

electricjones:
Sounds great. FYI Orient provides two methods to delete nodes. DELETE VERTEX will manage edges and DELETE FROM will not maintain the integrity of the graph. I use DELETE VERTEX.

So, once we add edges to the mix, spider will just use transactions by default when dropping any edge or vertex. In truth, transactions will be used quite a bit in the next iteration.

PR #68, opened 2015-08-23
made multiple code corrections for scrutinizer

dmill-bz:
You might want to have a quick look at this before we merge. All the tests pass and this will clean stuff up with scrutinizer and inspector though. 
(some corrections need more thought or will come naturally as the software evolve)


electricjones:
Looks good to me

PR #71, opened 2015-08-30
Changed action methods to verb-like names.

smolinari:
I also changed set() to getSet() as this also wasn't properly meaning an action. In fact, someone not knowing better could have understood set() to be action method, as in a setter. 

I also updated the docs accordingly. 

Hopefully all is satisfactory.

Scott

### DIFF
--- a/docs/command-builder.md
+++ b/docs/command-builder.md
@@ -14,7 +14,7 @@ $person = $query
    ->select()
    ->from('person')
    ->where('name', 'vadas')
-   ->first();
+   ->getOne();
 
 echo $people->name; // 'Vadas'
 ```
@@ -89,7 +89,7 @@ $builder->select()->from('person')
     ->limit(20)
 
 $builder->select()->from('software')
-    ->first();
+    ->getOne();
 ```
 
 Order the results.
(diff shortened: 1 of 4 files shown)


dmill-bz:
Nice thanks for this!! I noticed the following:

Theres a potential bug we can recreate like this:

``` php
$query = new Query(/** properties here **/);
$query->execute(new Command('script1','language')); // executes properly
$query->select()->getAll(); // executes script 1 again instead of a find all.
```

We should add this as a test to the suit and make sure it passes

smolinari:
Hmm.. good point. Probably a stupid question, but what would be the proper behavior? 

Scott

dmill-bz:
The query 'execution/dispatch' should probably reset the `Query` to prepare for a new query. I don't think this is something we should put forward in the documentation, it would just be nice to have that behavior as an added bonus (it's also a little more intuitive).

electricjones:
Originally, the `dispatch()` method reset the `CommandBag`. That said, there should probably exist a `reset()` method that will create a new `CommandBag` and reset any other parementers that may creep up as time goes gy.

This looks great @smolinari. Thanks.

smolinari:
No problem. Glad I could help and it is ok. I found a couple of smaller mistakes. However, I also played around a lot with other stuff. I have to revert some changes. Then I'll commit the few small ones, ok?

Will add the new test and changes as PommeVerde suggested? Or if you give me a tip on what to do, I can try it too.

Scott

dmill-bz:
I think it would be good to have `$command` be local to the dispatch method. Rather than a class property. It doesn't make that much sense for the `Query` class to hold the `Command` object.

smolinari:
Ok. I can change that. Probably a noob mistake. 

Scott

smolinari:
Ok. I've hopefully reverted all the stuff I was playing around with. LOL! 

Scott

PR #76, opened 2015-09-01
0.3.1

electricjones:
Fixes a few bugs:
- Neo4j passes consistently. Added ordering to queries in Query test
- OrientSql Command Processor uses `orderBy` constants correctly
- Several minor scrutinizer fixes.

I haven't come across any other bugs that would warrant being released before 0.4. Is there anything else that needs to be in 0.3.1?

After I merge and release, I will merge the updates into develop and down the tree. Be sure to `pull` changes.


electricjones:
Okay, I'm gonna release this for now. If we come across anything else, there are plenty more numbers :)

Issue #46, opened 2015-08-14
Minor refactors


electricjones:
A bucket list of small things to refactor as the project grows.
- [x] Split the current `where` logic between `Builder` and `BaseBuilder`. `BaseBuilder` should be simpler. PR #77 
- [x] The current implementation of converting Builder to Command in the drivers is verbose and an ugly copy/paste. We can change this once we have the other processors up and running, we'll just create a method in AbstractDriver to handle this. Should make the code much cleaner.
- [x] When inserting multiple records, the Orient SQL Processor method to create the insert data has way too many foreach loops.
- [x] Refactor OrientDriver to use SQL Batch in every case. This will also free us to simplify `DriverInterface` Commit 8a21b1a947b875779d876897ca0942825dd8b037
- [x] Run through some of the issues reported by scrutinizer and Insights. See the readme badges for details.
- [x] Add branches documentation to contributing documentation

Issue #23, opened 2015-07-28
Include value binding features to DriverInterface


dmill-bz:
Binding values to queries in order to prevent injections is going to be a common feature of all databases and driver clients. We should add functionality around these in `DriverInterface`

Something along the lines of:

``` php
/**
 * @return void
 */
public function bindValue(string $name, $value);

/**
 * @return void
 */
public function bindValues(array $binds);

/**
 * @return array (key/value)
 */
public function getBinds();
```

The discussion around binds can be much larger. We will need to settle on a consistent bindpattern (`{var}`, `B_VAR`, etc.) for `Command` and then it would be the `CommandProcessor`'s job to correctly format the binds to whichever language and pass them to the driver.

electricjones:
It should be possible to use the diver without the CommandProcessor if you know what you are doing.

To that end, what if the `Command` holds the script, the script language, and the values to bind? The driver then, when `executeXCommand()` is fired, binds the values to the statement before sending it along.

``` php
$driver = new OrientDriver($credentials)

$values = [
    'name' => 'Malcom Reynolds',
    'ship' => 'Serenity'
];

$command = new Command("INSERT INTO Profile SET {{values}}", $values);
// or $command->setValues($values);
```

The `Driver` then receives the `Command` and performs the correct cleanup and regex replacements.

``` php
public function executeWriteCommand($command) {
    $cleanCommand = // get the values from the Command, sanitize them, and regex replace
    // Becomes INSERT INTO Profile SET name = 'Malcom Reynolds', ship= 'Serenity'
    $response = // send it all to the database
}
```

I don't know if I like this, though, because it gives the driver responsibility in writing query statements. 

It may be an _additional_ option to binding in the `Driver` itself.

dmill-bz:
I've got a counter-proposal that is almost the same thing as what you suggested.
It works on the basis that:
- Every `Processor` should have functionality around cleaning and regex replacements of binds. Or at least extend a class that does.
- Knowledge of the format of binds for it's language. `{{value}}` | `:value` | etc. (we will have to settle on something if there is no specific format for a language)
- Return a `new Command('query string', 'language', ['bindKey' => 'bindValue'])`

With this there are two scenarios:
- The `Driver` has a specific tunnel for binds (ie: the server handles the bindings and ensures correct escaping/injection-protection). In which case the `Driver` will just add the binds separately to the request:
  
  ``` php
  $driver = new GremlinDriver($credentials)
  
  $values = [
[...]

PR #77, opened 2015-09-01
moves where to Builder, and simplifies constraints in BaseBuilder #46

electricjones:
This may require a bit of discussion. I split the logic from `where` between `Builder` and `BaseBuilder` so its more inline with the rest of the system. That required me to rename `BaseBuilder::where()` to `BaseBuilder::constrain()`.

The `Builder` API is identical and all tests pass.

The `BaseBuilder` api now ONLY accepts an array. This array may be a single or multiple full constraints. By full constraint I mean `[field, operator, value, conjunction]`. And you must use constants for operator and conjunction. `constrain()` tests that there are four elements in a constrain array and that operator and conjunction are both integers.

See tests/Unit/Commands/Builders/BaseBuilder/RetrieveTest.php for an example.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -19,11 +19,11 @@
             <directory suffix=".php">src/</directory>
         </whitelist>
     </filter>
-    <logging>
-        <log type="tap" target="build/report.tap"/>
-        <log type="junit" target="build/report.junit.xml"/>
-        <log type="coverage-html" target="build/coverage" charset="UTF-8" yui="true" highlight="true"/>
-        <log type="coverage-text" target="build/coverage.txt"/>
-        <log type="coverage-clover" target="build/logs/clover.xml"/>
-    </logging>
+    <!--<logging>-->
+        <!--<log type="tap" target="build/report.tap"/>-->
+        <!--<log type="junit" target="build/report.junit.xml"/>-->
+        <!--<log type="coverage-html" target="build/coverage" charset="UTF-8" yui="true" highlight="true"/>-->
+        <!--<log type="coverage-text" target="build/coverage.txt"/>-->
+        <!--<log type="coverage-clover" target="build/logs/clover.xml"/>-->
+    <!--</logging>-->
 </phpunit>
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -19,11 +19,11 @@
             <directory suffix=".php">src/</directory>
         </whitelist>
     </filter>
-    <!--<logging>-->
-        <!--<log type="tap" target="build/report.tap"/>-->
-        <!--<log type="junit" target="build/report.junit.xml"/>-->
-        <!--<log type="coverage-html" target="build/coverage" charset="UTF-8" yui="true" highlight="true"/>-->
-        <!--<log type="coverage-text" target="build/coverage.txt"/>-->
-        <!--<log type="coverage-clover" target="build/logs/clover.xml"/>-->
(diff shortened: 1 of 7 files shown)


dmill-bz:
I'll have a look at this sometime tonight. need to run off.

smolinari:
Maybe validateWhere() could be a better method name, since that is all the method is doing? 

Scott

electricjones:
It doesn't just validate the constraint, it actually adds the constraint to the CommandBag. This is the only method that actually adds anything to the command bag. Builder::where() just outputs a valid constraint and then hands it over to the BaseBuilder.

So, while I'm not married to `constrain()`, I'm not sure what is better.

dmill-bz:
why not just call it `where()`? I don't like the idea of having these extra methods lingering around in the api (even if protected).
you can then just have `Builder::where()` call `parent::where()` ?

electricjones:
That was my first reaction, but the signatures are different and it throws a fatal error

dmill-bz:
oh how interesting, didn't see that coming.

electricjones:
Me neither.

dmill-bz:
Ok, I got reading a bit around and it's a compiler thing. Which is a real bummer for us. It's a shame php doesn't allow for multiple signatures. Anyways, how about naming it `internalWhere()`?

electricjones:
That works for me.

electricjones:
At least for now.

electricjones:
I'm heading out for a while. If you don't have any objections, please merge this. Or, feel free to push a commit and merge.

dmill-bz:
Oh I meant to ask. was there a reason behind removing the clover.xml generation?

electricjones:
Oh, my bad. I disable it while I'm working (to save time) and I forgot to re-enable it.

dmill-bz:
looks good :)

smolinari:
This is me learning (and probably being picky again :smile: and to be honest, I really don't want to be a fifth wheel or a hindrance), but from what you said Chris, that my suggested `validateWhere()` (now `internalWhere()`) does more than just validate, I was thinking, is BaseBuilder then breaking SRP? Because, theoretically the method name would actually need to be `validateWhereAndAddToBag()` and that got me thinking, maybe there needs to be a `BagValidator` class? If there would be one, then a good bit of the logic could be removed out of the `Builder` and `BaseBuilder` classes and they would only be responsible for building the command bag and no longer be responsible for also validating any parameters. I also noticed not every parameter is checked for validity either. For instance, `$limit` should be an integer, if I am not mistaken. If I give `limit()` a string parameter, it would probably break Spider. (I haven't tested, so just guessing). 

What do you think? Is this a good challenge for me again, since I brought it up? Or, am I wrong and taking SRP too far, because, what reasons for changes are there for the Builders? It is pretty singular, or is it?

Scott

dmill-bz:
I'm not all that bothered with us not validating the data. If you where to write a query using brackets around some info and that happened to be the wrong type, the server would let you know by throwing an error. If you place a string in limit() you get the same result. This doesn't strike me as all that bad.

In some cases you might even want to pass a string to limit (depending on the backend storage and how their limit works) so again it's just more flexible. In all cases you should get an explanatory error, it'll just be from the server and not spider

smolinari:
Ok. Is there some rule of thumb, as to when parameters offered by the client should be validated?

Scott

dmill-bz:
Well they should all be escaped to avoid injections. But that's something that will come with the implementation of issue #23

Issue #81, opened 2015-09-03
Multiple CRUD operations per query.


dmill-bz:
This is an issue to discuss the implementation of CRU and RD scenarios.

Here are some scenarios we need to cover:
1. Create (C) two vertices and an edge between them
2. Find (R) existing vertices and create (C) an edge between them
3. Find (R) existing vertices and create (C) an edge between them then update that edge with properties (U)
4. Create (C) elements and update (U) them with data. 
5. Find (R) and delete (D) elements (should already work but leaving here incase we decide to make it work in a similar fashion to the above)

Lets try and suggest ways of writing the _declarative_ queries for each case:
## 1

``` php
(new BaseBuilder())->create([   
   [TYPE => VERTEX, LABEL => 'person', "name" => 'what'],
   [TYPE => VERTEX, LABEL => 'person', "name" => 'ever'],
   [
      TYPE => EDGE,
      LABEL => 'friend',
      INV => (new BaseBuilder())->create([[TYPE => VERTEX, LABEL => 'person', "name" => 'michael']]),
      OUTV => (new BaseBuilder())->create([[TYPE => VERTEX, LABEL => 'person', "name" => 'dylan']])
   ]
]);
```

I added some extra vertices to show how we could possibly batch
## 2

``` php
(new BaseBuilder())->create([
   [
      TYPE => EDGE,
      LABEL => 'friend',
      INV => (new BaseBuilder())->select()->where(["name", EQUALS, "michael", AND]),
      OUTV =>  (new BaseBuilder())->select()->where(["name", EQUALS, "dylan", AND])
   ]
]);
```
## 3

``` php
(new BaseBuilder())->create([
   [
      TYPE => EDGE,
      LABEL => 'friend',
      INV => (new BaseBuilder())->select()->where(["name", EQUALS, "michael", AND]),
      OUTV =>  (new BaseBuilder())->select()->where(["name", EQUALS, "dylan", AND])
   ]
])
->update()->withData(['weight' => 15]);
```

What to do here if we batch created? The above would apply the data to all of them? How to apply the data selectively?
## 4

Same as above really. With all the same questions.
## 5

Currently we use:

``` php
(new BaseBuilder)->delete()->where(['name', EQUAL, 'dylan', AND]);
```

Don't think this needs to change.

What are your suggestions?

electricjones:
This is basically exactly what I was thinking. A few thoughts:

## Data and fields

I would prefer to use the term 'data' consistently. So, for **1**:

``` php
[...]

Issue #29, opened 2015-08-02
NotSupportedException vs Not supported warning.


dmill-bz:
I just wanted to bring this up for discussion.

The idea is that some graph DBs are going to support some features while others wont. Examples range from tree functionality to transactions to indexing capabilities.
Now if every DB Driver or Schema implementation throws exceptions on unsupported features, this means that ultimately, switching backends will not be possible without a wave of exceptions being thrown, and code refactor. 

I feel it's our duty to make the transition from one backend to another as easy as possible. Hence the introduction of warnings.

The idea is that if you are using transactions in your app with a transaction enabled GraphDB, and then you decide to switch to a transaction disabled GraphDB, the system will emulate transactions but throw a warning (non thread breaking). Rather than outright throw a thread breaking exception.

Same could be done with creating indexes if it isn't supported, the system just ignores the index creation and throws a warning. (queries etc.. still work they just don't use index optimization)

Some cases like tree however will need to continue throwing exceptions as it would be a potential app breaker.

Thoughts?

electricjones:
I like the idea, but its implementation may be difficult. Even warnings can be a problem in production code. No system is going to be 100% db swappable. Instead, lets try to make it as painless as possible. Say (in most cases), no more than 2hrs work, or whatever.

We will eventually have a top level `Spider` which will hold configuration and be a factory of sorts. What if we gave the option for three levels of warnings:
1. Fatal - every not supported is fatal.
2. Quiet - we produce warnings, probably by catching the exceptions and raising warnings
3. Silent - when possible, no output is rendered. Only logged using the logger of their choice.

Which also raises the issue of logging and caching with third-party packages. Simple interfaces for logging, events, and caching may be helpful. FIGG is working on these, so we can implement them at a time in the future.

There is also the issue of multi-model databases that need to be considered, though in a different issue. For instance, OrientDB has polymorphism in its vertex types (labels) and document db features. There should exist a way for third-party packages to _extend_ Spider for these db-specific features.

dmill-bz:
I agree that all sounds good. As For OrientDB, I know that they are currently working on a TP3 implementation. I'd be curious to see how that works out before moving forward as it may shed some lighting on how to handle these scenarios.

electricjones:
I am going to add this configuration into the top-level api. Issue #22

electricjones:
Moved into #22. Closing conversation here.

Issue #22, opened 2015-07-28
Top Level Spider API


electricjones:
The `Connections\Manager` served its purpose once, but now that so much more is going on, it needs to be refactored into the top-level `Spider\Spider` as a convenient place to store all configs, create and cache whatever objects (drivers, builders, connections, processors, etc), and hold helpers and utilities.

These are all rough thoughts

``` php
// You may set global configuration
Spider\Spider::setup([
  'default' => 'gremlin',
  'connections' => [ // I don't like this term. This is a package of driver, processor, and credentials
    'gremlin' => [
        'driver' => 'Full\Class\Driver',
        'processor' => 'Full\Class\Processor', // or use convention, which I think I'd prefer
        'hostname' => 'localhost' // and the rest of configurations    
    ],
    'gremlin-two' => [
        'driver' => 'Full\Class\Driver',
        'processor' => 'A\Different\Full\Class\Processor'
        'hostname' => 'remote' // and the rest of configurations    
    ],
    'orient' => [
        'driver' => 'Full\Class\Driver',
        'processor' => 'Processor'
        'hostname' => 'remote' // and the rest of configurations    
    ],
    // etc
])

// And then make a new instance
$spider = Spider::make(); // default by default
$spider = Spider::make($connection);
$spider = new Spider($config);

// The returned Spider object extends the CommandBuilder and contains the correct connection, so you may use the querybuilder directly
$spider->select()->where()->get();

// Or return its pieces
$spider->getDriver();
$spider->getProcessor();

// Or make new instances of components (eventually includes)
$spider->connection(); // returns a new connection from config, alias, or empty
$spider->graph(); // returns a new empty graph (model) with current connection
$spider->schema(); // returns a schema builder with current connection
$spider->make(); // makes a new Spider with current connection
$spider->reset(); // destorys current spider and makes a new one
```

electricjones:
On this note, we may be able to lose `Connection` all together. All it really does now is pass things through to the driver. I had included it to leave some wiggle room for the future, but I don't remember what I had in mind and now I cant think of a reason to have a middle-man connection.

dmill-bz:
Connection could handle Transactions (simulating nested transactions), query caching, load balancing in some cases. etc.. I think we should keep it. 
I'll have a bit more time to look at the PR tomorrow :)

electricjones:
Sounds great.

electricjones:
Have begun a very work-in-progress top-level Spider api in branch `spider-api` with the following goals:
[...]

Issue #19, opened 2015-07-23
Add Traversals and nested queries to CommandBuilder


electricjones:
Once the basic `CommandBuilder` is up and running, it's time to play with relationships, edges, and graphs :)

Following a declarative paradigm, the user should _describe_ what all they want and have it returned.
### Examples

Here is a query that returns a nested set of
- The first user named "michael"
- michael's friends who have been friends longer than 3 years, only selecting their username and country

``` php
$query->select() // indicates read, select *
      ->from('users')
      ->where('username', 'michael')
      ->limit(1)
      ->where([
          'out' => ['where'=>[['label', 'friends'], ['years', '>', 3]]], // alias to outE(). Also in() and inE()
          'only' => [['username', 'country']],
          'store' => 'b'
          ]
      ])
      ->return('b') // return() by itself returns last stored
      ->tree();
```

Returns

``` js
{
   'username': 'michael',
   'country': 'USA',
   'nickname': '...' // and whatever other properties users have
   'friends': [
       {
          'username': 'eliza',
          'country': 'Spain'
       },
       {
          'username': 'jon'
          'country': 'England'
       },
       // etc
   ]
}
```

dmill-bz:
I'm not sure about calling the link "friends" I think this is something that will need to be viewed while discussing how Tree works.

dmill-bz:
Decided I'd talk about `Tree` here.

There are several levels to this.

**`Response` layer**

Returning the following:

```
'friends': [
       {
          'username': 'eliza',
          'country': 'Spain'
       },
       {
          'username': 'jon'
          'country': 'England'
       },
       // etc
   ]
```

Could be an issue for several reasons. The main ones being that there is no guaranty the user will input the Edge label. They might just want to follow all outgoing edges. Or they might specify multiple labels to follow. 

I feel that the cleanest way to implement this is unfortunately a little verbose. And it would be this : 

```
{
   'object' : 
   {
      'username': 'michael',
      'country': 'USA',
      'nickname': '...' // and whatever other properties users have
   },
   'children':
   {
       {
          'object':
          {
             'username': 'eliza',
             'country': 'Spain'
          },
          'children':{}
       },
       {
          'object':
          {
             'username': 'john',
             'country': 'England'
          },
          'children':{}
       },
       // etc
   }
}
```

The "object" would essentially be a `Spider\Base\Collection`.

**`Model` layer**

[...]

PR #82, opened 2015-09-03
Adds top level spider api with ioc, integrations, and configuration

electricjones:
This is a continuation of discussion #22. There is no rush here. #81 and #19 are probably the priority for now. Finally, I'm not married to any of these names or terms.

 _This PR is NOT ready to merge_. 

 This is a full and passing working draft, but could still use some cleanup, even if nothing changes structurally. Also note that I haven't updated the docs yet, but am going to detail everything here. I will update this comment as we come to decisions and it will turn into documentation before merging.

 Sorry, this is pretty long, but here we go.

---
### Goals for the Top-Level API
- [ ] Be able to configure spider in one location and then re-use that configuration
- [ ] Have sensible, cascading defaults for those configurations
- [ ] Pass that configuration to child objects, if using the Spider as a factory
- [ ] Allow user to handle `NotSupportedExceptions` #29 
- [ ] Include factories for query builders, connections, and more in the future
- [ ] Introduce a _light_ ioc container only into the top-level spider. Not required by components.
- [ ] Introduce an replaceable, optional event system. Also serves as an example for future integrations (caching, logging, etc).
  ### How Spider works
  - [ ] Configure a static via `Spider::setup()`
- [ ] Instantiate a new `$spider = Spider::make()` from default or other set connection
- [ ] That `$spider` extends the query builder `$spider->select->etc()`
- [ ] That `$spider` also acts as a factory:
- [ ] `$spider->connection()`, `$spider->querybuilder()` returns new, configured objects
- [ ] Anything used through `$spider` is passed optional configurations and the event dispatcher
- [ ] Those components use `ConfigurableTrait` and `UsesEventsTrait` which manage events and configuration. Note that eventing is optional. If the Connection Manager fires an event with no dispatcher set, that is fine
  ### Configuration
  
  Taking advice from @smolinari
- [ ] User MUST set a _default connection_ and at least one valid _connection_
- [ ] User MAY replace the default _ioc container_ with one that implements `Michaels\Manager\Contracts\IocContainerInterface` which only has a `fetch()` method
- [ ] User MAY replace the default _event dispatcher_ with any that implements `Spider\Integrations\Events\DispatcherInterface` (more on that below)
- [ ] User MAY define how _unsupported errors_ are handled (more on that below)
- [ ] User MAY use any _combination_ of the above as defaults will be provided
  
  ``` php
  Spider::setup([
     'connections' => [
         'default' => 'neo',
         'neo' => [
             'driver' => 'neo4j'
             // The rest of the parameters
         ],
     ],
     'integrations' => [ // optional
         'events' => $dispatcher // many ways to set dispatcher below
     ],
     'errors' => [
         'not_supported' => 'silent' // defaults to fatal, more below
     ]
  ], $optionalIocContainer);
  ```
  ### Instantiation and Factories
  
  Once the static instance is `setup`, the user can create instances
  
  ``` php
  $spider = Spider::make(); // default connection
  $spider = Spider::make('another_connection');
  ```
  
  Once an instance is made, you can also create components:
  
  ``` php
  $qb = $spider->querybuilder();
  $con = $spider->connection();
  // Or
  $qb = $spider->querybuilder('another_connection');
  $con = $spider->connection('another_connection');
  ```
  
  There are also a few managing methods
  
  ``` php
  $spider->getDriver()
  $spider->getConnection();
  $spider->getEventDispatcher();
  $spider->addConnection();
  ```
  - [ ] **Question 1**: Do we want a way to get the _same_ spider instance each time? `Spider::getInstance()`?
  ### Dependency Injection and Integrations
  
  This branch uses a modified version of my `michaels/data-manager` as a lightweight ioc container. I will release that version before this is merged.
  
  To avoid coupling the code too tightly, the IoC container never actually leaves the top level `$spider`. Instead, it only _holds_ objects and injects them into the constructors of the various components (Connections Manager, Query, Drivers, etc). This means that if someone is using the pieces independently of spider, they do not need to worry about the container, only set the dependencies _if they want to_.
  
  At the moment only events are used, but eventually I can picture caching, logging, and maybe profiling, among potential others.
  
  You can also set your own ioc container (say the one your framework uses), so long as it implements the interface.
  - [ ] **Question 2**: Would it be better to use duck-typing here? It may be difficult to wrap a frameworks container with a contract class. Is `fetch()` not standard enough?
  
  The supplied IoC container allows you to set dependencies in four ways:
  1. As a _string_ class name.
  2. As an already instantiated _object_
  3. As a closure _factory_
  4. As an instance of a _container_ with fetch. In this case, when the dep is needed, the library calls `$container->fetch('alias')`
  
  ``` php
  Spider::setup([
     'integrations' => [
         'events' => 'Some\Class\Name',
         'events' => $dispatcher, // or new Dispatcher()
         'events' => function () {
            return new Dispatcher();
         }
         'events' => $container, // with a registered dep called 'events'
     ]
  ]);
  ```
  ### Using Events
  
  At the moment, only the connection manager uses events. This was just done as a workflow example. Before this is merged (if we go forward with events now), we may need to move it into the rest of the system.
  
  To setup a component to use events, simply
  1. Include the trait
  2. Modify the constructor to accept and set a dispatcher
  3. Fire off events, confident that with no dispatcher, nothing will happen
  
  ``` php
  class SomeClass
  {
     use UsesEventsTrait;
  
     public function __construct(DispatcherInterface $events = null)
     {
         // Optional Event Dispatcher
         $this->setDispatcher($events);
     }
  
     public function someMethod()
     {
            $this->emit('my.event');
     }
  }
  ```
  
  You may also add events. This is shown in the tests.
  
  ``` php
  $events = [];
  $spider->getEventDispatcher()->addListener(
     'connections.manager.before_make',
     function () use (&$events) {
         $events[] = 'fired-before_make';
     });
  
  $spider->getEventDispatcher()->addListener(
     'connections.manager.after_make',
     function () use (&$events) {
         $events[] = 'fired-after_make';
     });
  
  $spider->connection(); // will fire both events
  
  $this->assertEquals('fired-before_make', $events[0], "failed to fire before");
  $this->assertEquals('fired-after_make', $events[1], 'failed to fire after');
  ```
  ## Configurable and NotSupported Traits
  
  This also introduces two new traits that add functionality to component parts.
  
  The `ConfigurableTrait` allows a component to hold optional configuration in a predictable way. It has an associated interface.
  - When instantiating the component, pass an optional `$configManager`, which can be an instance of `Michaels\Manager\Manager` or an array of configurations.
  - In the constructor, `setConfigMangaer()`. If not using config, then an empty manager is created
  - In the class, use `$this->config()->get('whatever')` to return the manager and get the item. See michaels/manager for all the options
  
  The `NotSupportedTrait` allows a component to declare a feature not supported and the trait handles how that issue is raise via exception or warning or silent. 
  
  The `NotSupportedTrait` requires the `ConfigurableTrait`.
  - `use Spider\Base\ConfigurableTrait`
  - `use Spider\Base\NotSupportedTrait`
  - When something is unsupported, simply `$this->notSupported($message)` and the trait will use the configuration to handle the error.
    - `'fatal'` throws a `NotSupportedException`
    - `'quiet'` raises an `E_USER_WARNING`
    - `'silent'` does nothing (this will probably lead to another failure down the chain.
    - defaults to fatal. We can change these names and add whatever we wish.
  
  Take a look at the `AbstractDriver` and one of the Driver implementations to see a full rundown.
  ### Open Questions and Tasks
  - [ ] **Task 1**: Split testing. Right now, all testing occurs in `Integrations/SpiderTest`, but should probably be split.
- [ ] **Task 2**: Better support of Event Testing, probably in its own test
- [ ] **Question 3**: Right now, it is possible to configure statically and use `make()` **or** to configure on an instance using `configure()` **or** to pass the configuration into the constructor. This seems flexible but obscure. We should probably settle on a 1 or two solid methods.
- [ ] **Question 4**: Is there a way to get rid of the `ThrowsNotSupportedStub` which is only used in the `ThrowsNotSupportedTest`?
- [ ] **Task 3**: When testing whether Spider passes not supported config and such, it relies on `getTree()` being unsupported, which it will not be forever.
- [ ] **Task 4**: Turn this PR into documentation and update docblocs.
- [ ] **Task 5**: Integrate Events, NotSupported, and Configuration throughout the system?

Of course, any feedback is welcome.


dmill-bz:
Just waned to let you know that I will eventually get around taking a look at this PR

electricjones:
That's fine. I figure this would be the last feature we merge into 0.4 before we bug scrub and cleanup for the release.

electricjones:
I simplified the way defaults are loaded and changed the `data-manager` dependency to the `develop` branch. I will soon release `0.8.5` and the dependency will change again.

Please `composer update`

PR #84, opened 2015-09-04
set of changes for the scrutinizer configuration

dmill-bz:
_Not ready to be merged_
Setting this here to run it against Scrutinizer. 
Should I also go ahead and change travis to check `PHP 5.6` only as well?

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -7,6 +7,25 @@ checks:
     php:
         code_rating: true
         duplication: true
+        use_self_instead_of_fqcn: true
+        uppercase_constants: true
+        return_doc_comments: true
+        return_doc_comment_if_not_inferrable: true
+        properties_in_camelcaps: true
+        parameters_in_camelcaps: true
+        param_doc_comment_if_not_inferrable: true
+        no_goto: true
+        newline_at_end_of_file: true
+        function_in_camel_caps: true
+        fix_use_statements:
+            remove_unused: true
+            preserve_multiple: false
+            preserve_blanklines: false
+            order_alphabetically: true
+        fix_line_ending: true
+        classes_in_camel_caps: true
+        avoid_perl_style_comments: true
+        avoid_multiple_statements_on_same_line: true
 
 tools:
     php_code_sniffer:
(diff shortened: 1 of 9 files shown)


electricjones:
Yeah, let's just do `5.6`, `7`, and `hhvm`

dmill-bz:
Ok so I've made it so that **PSR-2** is the standard used for checking the code. I'm not sure how well this is being carried out as I still see some PSR-2 standards in the "not enabled" checks. This might be because I configured code sniffer directly. Anyways, It won't hurt to have some redundant checks so on top of this here's what I suggest we add as extra checks (check them off if you agree):
- [x] Check that `self` is used in place of fully qualified names
- [x] Check that constants are in `UPPER_CASE`.
- [x] Check the types of `@return` doc comments 
- [x] Check whether the return type is inferable and ask for a `@return` comment if not
- [x] Check that properties are named in camelCaps format (_I don't know if this interferes with private `_` prefixed properties. We could always disable it later._)
- [x] Check that parameters are named in camelCaps format
- [x] Check whether parameter types are inferable and ask for a `@param` comment if not (should mostly cover missing `@param` and not suggest what they should be
- [x] Check that optional function parameters are defined at the end 
- [x] Check that files end with a new line character
- [x] Check that function names are in camelCaps format 
- [x] Fix use statements 
- [x] Check that class names are in CamelCaps format
- [x] Check that Perl-style comments (# Some comment) are not used
- [x] Check there are not multiple statements on the same line (e.g.: `blah;blah;`)

Again, some of these should already be covered by **PSR-2** checks, no harm in making them redundant.

dmill-bz:
Oh and I place 7.0 as an allowed failure (as some of the dependencies fail).

electricjones:
Looks great. Should raise our score by default. Thanks for the work.

dmill-bz:
I'm not sure the score will increase. Usually the score is related to cyclomatic complexity, potential bugs, structure of classes (abstract, or if there are too many private/protected methods, etc.). So it should stay roughly the same. But at least we won't have constant unrelated nagging from scrutinizer.

dmill-bz:
ok all tests pass I'm going to merge this

smolinari:
:+1: 

Scott

Issue #53, opened 2015-08-17
Builder and Query Responsibilities


electricjones:
A recap of the conversation started in #22:

---

_@PommeVerte_ Aaaaaand while I'm at it. I've been thinking that we should provide a `TraversalBuilder` that extends `BaseBuilder` with a traversal syntax. I know we originally wanted to add sugar to Builder for this but thinking about it, it will be too complicated to do something clean (overlap and conflicts between what some methods would do). I have a couple of ideas here as well.

---

_@chrismichaels84_ Oh I didn't think about `Query`. `Builder` and `TraversalBuilder` would have been two leaves (it's not possible to have it function properly any other way as all the methods would function differently in populating the Bag) so maybe that would imply a different Query class as well.

I'd much rather add the basic traversal features to `BaseBuilder` so it can:
1. CRUD vertices and edges with labels and types (already implemented)
2. Manage relationships between vertices (`$builder->addEdge($from, $two, $label, $properties` or something)
3. Perform basic relationship queries (get all friends of user 3 who have been friends more than 2 years, or get all posts tagged as new and important, etc.)
4. Perform powerful, if verbose queries, sub queries, and traversals.
   ________________
   _@PommeVerte_ I think a lot of what you mention would probably be better on the `Model` layer. It's an abstraction on the logic (query construction) and not really query sugar. 

Also all methods would conflict.

``` php
$builder->out()->out()->where().
```

The `where` here would have to function completely differently to the current where that just adds stuff to the bag since this where would need to apply to some sub query. Actually a lot would have to be happening in the above. Anywho I don't see a `where()` or many other methods working for both scenarios.
Also what happens if you use `out()` in a declarative scenario? Sounds like a headache waiting to happen. I don't think there's much harm in providing two builders, two queries is a bit more of an issue though.

---

_@chrismichaels84_ I don't mind two builders. Its the two queries that rub me the wrong way. It complicates things for the user having to choose between or carry two different objects around in their code. This may be simplified in the top level api like

``` php
$spider->traverse($start)->out()->out()->with();
$spider->query()->select()->from();
```

But that still seems unwieldy.

Having two builders also makes it difficult for a higher abstraction (model) to know which builder to use without being explicitly told. At the end of the day (top-level), the user shouldn't have to make that choice. They should declare what they want from a `model` or `spider` and it happens.

If we use subselects (as orient sql does), then most of the above is possible (with the exception, maybe, of number 4. I'd have to ask around).

---

electricjones:
[...]

PR #85, opened 2015-09-15
Multiple CRUD operations per query

electricjones:
I finally had a couple hours, so I began work on multiple CRUD operations in a single command. The approach is a little different that we discussed, but it meets all of our goals, I think.

_This is not ready to merge!_ This is only the first draft. This PR currently breaks `Builder` and `Query` tests because I don't want to update them until this is decided upon. It does not affect any other tests and (of course) `BaseBuilder` tests have been updated and pass.

This PR begins to address the features required by #81, which is required before we dive into #19 and #53.

---

Here are some scenarios we need to cover:
     1. Create (C) two vertices and an edge between them
     2. Find (R) existing vertices and create (C) an edge between them
     3. Find (R) existing vertices and create (C) an edge between them then update that edge with properties (U)
     4. Create (C) elements and update (U) them with data.
     5. Find (R) and delete (D) elements (should already work but leaving here incase we decide to make it work in a similar fashion to the above)

A quick review of the goals from #81:
- [ ] Allow for nested-commands that are instances of `BaseBuilder` or `Bag`s 
- [ ] Allow for _declarative_ creation of vertices and edges.
- [ ] Switch to a syntax (for `BaseBuilder`) that is `$builder->select()->from()->delete()`
- [ ] Simplify `BaseBuilder` responsibility and prefix methods with `internal`
- [ ] Allow for aliasing operations via `set()` and `get()`
- [ ] Make the processors do the work

So far, this PR should cover all of the above, but only in the scope of `BaseBuilder`. I haven't touched processors, other builders, or drivers. Also, the nested queries should work because `BaseBuilder` has no awareness of them, but there are no tests for it. Tests cover everything else.
## General
- [ ] Prefixed CRUD to `internalCreate()`, `internalRetrieve()`, `internalUpdate()`, `internalDelete()`, `interalProjections()`, and `internalWhere()`
- [ ] changed `path()` and `tree()` to `setAsPath()` and `setAsTree()`
- [ ] Removed `data()` and `type()`
- [ ] left methods that aren't overridden the same (`limit()`, etc)
- [ ] I also cleaned up `Bag` a little, dropping `createCount` for instance. I think we can drop `target` as well. 
## Multiple Operations

I did end up taking a different approach that we discussed. Instead of having `Bag::$create` etc, I opted for `BaseBuilder` creating an _array_ of CommandBags. Ordering is strictly kept and a new Bag is only added to the stack when one of the CRUD methods is called. I did this because it seemed simpler, more flexible, and more intuitive in controlling what _order_ the operations are executed in.

**A breakdown of what's going on** from the scenarios above:

---
## 1

``` php
(new BaseBuilder())->internalCreate([   
   [TYPE => VERTEX, LABEL => 'person', "name" => 'what'],
   [TYPE => VERTEX, LABEL => 'person', "name" => 'ever'],
   [
      TYPE => EDGE,
      LABEL => 'friend',
      INV => (new BaseBuilder())->internalCreate([[TYPE => VERTEX, LABEL => 'person', "name" => 'michael']]),
      OUTV => (new BaseBuilder())->internalCreate([[TYPE => VERTEX, LABEL => 'person', "name" => 'dylan']])
   ]
]);
```
## 2

``` php
(new BaseBuilder())->internalCreate([
   [
      TYPE => EDGE,
      LABEL => 'friend',
      INV => (new BaseBuilder())->internalRetrieve()->internalWhere(["name", EQUALS, "michael", AND]),
      OUTV =>  (new BaseBuilder())->internalRetrieve()->internalWhere(["name", EQUALS, "dylan", AND])
   ]
]);
```
## 3

``` php
(new BaseBuilder())->internalCreate([
   [
      TYPE => EDGE,
      LABEL => 'friend',
      INV => (new BaseBuilder())->internalRetrieve()->internalWhere(["name", EQUALS, "michael", AND]),
      OUTV =>  (new BaseBuilder())->internalRetrieve()->internalWhere(["name", EQUALS, "dylan", AND])
   ]
])
->update(['weight' => 15]);
```

In this specific case, the processor would have to know to look at the previous operation in the batch to query which edges to update. _I don't like this_

A better way:

``` php
(new BaseBuilder())->internalCreate([
   [
      TYPE => EDGE,
      LABEL => 'friend',
      INV => (new BaseBuilder())->internalRetrieve()->internalWhere(["name", EQUALS, "michael", AND]),
      OUTV =>  (new BaseBuilder())->internalRetrieve()->internalWhere(["name", EQUALS, "dylan", AND])
   ],
   'weight' => 1.5
])
```
## 4

Same as above
## 5

``` php
$builder->internalRetrieve()->internalWhere()->delete();
```

In this case, again, the processor would have to know to use the previous operation to find which to delete. At least, this is what Orient would translate to

```
DELETE VERTEX (SELECT FROM V WHERE x = y)
```

or

```
begin batch
LET a = SELECT FROM V WHERE x = y
LET b = DELETE VERTEX $a
commit
```
## Setting aliases

``` php
$builder->internalCreate([
   TYPE => VERTEX,
   LABEL => 'person'
   name => 'michael'
])->set('michael');

$builder->internalCreate([
   TYPE => VERTEX,
   LABEL => 'person'
   name => 'dylan'
])->set('dylan');

$builder->internalCreate([
   TYPE => EDGE,
   LABEL => 'friend'
   weight=> 3
   INV => $builder->get('michael')
   OUTV => $builder->get('dylan')
]);
```

---

This is all still a rough draft, but let me know what you think. I will probably comb through this again tomorrow to see if I can clean it up a bit and make sure it all works as expected for edge cases.

### DIFF
--- a/src/Commands/Bag.php
+++ b/src/Commands/Bag.php
@@ -114,4 +114,7 @@ class Bag extends Object
     const ELEMENT_EDGE   = 510;
     const ELEMENT_LABEL  = 520;
     const ELEMENT_ID     = 530;
+    const ELEMENT_TYPE = 540;
+    const EDGE_INV = 550;
+    const EDGE_OUTV = 560;
 }
--- a/src/Commands/Bag.php
+++ b/src/Commands/Bag.php
@@ -42,9 +42,6 @@ class Bag extends Object
     /** @var array Data to be inserted/updated */
     public $data = [];
 
-    /** @var int How many records to create */
-    public $createCount = 0;
-
     /**
      * What do you want after an operation is complete?
      *
--- a/src/Commands/BaseBuilder.php
+++ b/src/Commands/BaseBuilder.php
@@ -158,17 +158,6 @@ class BaseBuilder
 
     /* Fluent methods for building queries */
     /**
-     * Set the type of the target in the current Command Bag
-     * @param $type
-     * @return $this
-     */
-    public function type($type)
-    {
-        $this->addToCurrentBag('target', $type);
-        return $this;
-    }
-
-    /**
      * Add data to the current command bag (for insert and update)
      * @param $property
      * @param null $value
--- a/src/Commands/BaseBuilder.php
+++ b/src/Commands/BaseBuilder.php
@@ -10,7 +10,7 @@ use Spider\Commands\Languages\ProcessorInterface;
  */
 class BaseBuilder
 {
-    /** @var Bag The CommandBag with command parameters */
+    /** @var array An array of CommandBags with command parameters */
     protected $bag;
 
     /** @var  ProcessorInterface The current language processor */
@@ -24,33 +24,27 @@ class BaseBuilder
      * With an optional language processor
      *
      * @param ProcessorInterface|null $processor
-     * @param Bag|null $bag
+     * @param array|null $bag
      */
     public function __construct(
         ProcessorInterface $processor = null,
-        Bag $bag = null
+        array $bag = null
     ) {
         $this->processor = $processor;
-        $this->bag = $bag ?: new Bag();
+        $this->bag = $bag ?: [];
     }
 
-    /* Fluent Methods for building queries */
+    /* Internal methods for building queries */
     /**
      * Add an `insert` clause to the current command bag
      * @param array $data
      * @return BaseBuilder
      */
-    public function insert(array $data)
+    public function internalCreate(array $data)
     {
-        $this->bag->command = Bag::COMMAND_CREATE;
-
-        if (isset($data[0]) && is_array($data[0])) {
-            $this->bag->createCount = count($data);
-        } else {
-            $this->bag->createCount = 1;
-        }
-
-        $this->bag->data += $data;
+        $this->addNewBag();
+        $this->addToCurrentBag('command', Bag::COMMAND_CREATE);
+        $this->addToCurrentBag('data', $data);
 
         return $this;
     }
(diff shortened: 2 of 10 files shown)


electricjones:
So, in this workflow, there are a couple of things the processors will need to be updated to do:
- [ ] Accept an array of `Bag`s.
- [ ] Look for `Bag` or `BaseBuilder` in certain places (like INV and OUTV)
- [ ] Recursively process those nested command bags
- [ ] Be intelligent about using the previous operation in the current operation in certain circumstances (like delete).

To do before this can be merged:
- [ ] Settle on method for multiple crud (array of bags or something else)
- [ ] Update processors
- [ ] update Builder and maybe Query with tests
- [ ] I don't think drivers will need an update, but take a look.
- [ ] Possibly create a `BagContainer` or something that implements `ArrayAccess` to be used instead of native array. This way we can do some helpers to manage pointers (`getPreviousBag()`, etc)
- [ ] documentation

dmill-bz:
I've only had a very quick read but I'll be more thorough later today. My first concern is that these two commands should yield the same internals:

``` php
(new BaseBuilder)->select()->where(/*...*/)->update(/*data*/);
```

``` php
(new BaseBuilder)->select()->update(/*data*/)->where(/*...*/);
```

This is a requirement to keep the declarative aspect of the builder (possibility to: scope, add some sugar,  etc.) and from a quick glance this PR breaks that? (Correct me if I'm wrong, I only took a glance at the code). Basically we should be able to put all the commands in any order and still yield the exact same output (as in the construct of what we send to the processor should be the same), that's what the original approach was supposed to cover. 
I have some other concerns but how we work around the above will define if they're valid or not.

electricjones:
You know, I just re-read the entire conversation and I must have been missing something. I think I just realized the difference between multiple operations inside a command and transaction support. I was treating this as transaction functionality at the builder level. Let me close this PR and take another stab at it. Should only take me an hour.

PR #93, opened 2015-10-08
fix all dependencies on semver

electricjones:
This will be 0.3.2 -- just a pr to make sure tests pass.

After merged, I will merge this into develop.

This keeps php 5.4. I'll commit to `develop` for 5.6 and then we can merge the changes down from there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,15 @@
 # Changelog
 All Notable changes to `Spider` will be documented in this file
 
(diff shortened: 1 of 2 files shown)


electricjones:
Well, it seems that the neo4j driver has stopped releasing updates, so we have two choices:
1. use an unfixed dependency
2. use a different library

Obviously, its easier to use an unfixed dependency, at least for now.

PR #98, opened 2015-11-21
Update gremlin driver

dmill-bz:
Updated the gremlin Driver to use the newer version of gremlin-php and set tests against gremlin-server 3.0.2.

If this passes the tests I think it's safe to merge without consensus.


dmill-bz:
wrong branch

PR #103, opened 2015-11-30
Update to run with ODB 2.1.6 and some smaller changes

smolinari:
This has a number of changes, so the ODB studio also works properly.

Scott

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
-composer.lock
 vendor
 build/
 docs/api/cache
+.vagrant
+vagrant*
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ build/
 docs/api/cache
 .vagrant
 vagrant*
+.idea/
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -1,4 +1,6 @@
-composer.lock
```
PommeVerte: this shouldn't be removed. instead it should be updated with `composer update <vendor>/<package>`

dmill-bz:
This looks good. Lets see what @chrismichaels84 has to say

smolinari:
Ok. I've also cleaned it up. I hope it is mergeable. :smile: 

Scott

electricjones:
@smolinari This looks good to me, but I can't test it right now. My internet isn't so great atm. I'll have a pull and run vagrant tomorrow just to be sure.

@PommeVerte, it would be nice if vagrant and Travis shared some bootstrapping. That  way we only have to change versions in one spot. I tried (for a few minutes) but had troubles with working directories. Any thoughts? Is this even worth the effort?

dmill-bz:
Yeah I think it's worth the effort. I didn't have a proper look at the vagrant build but I think a few things can be done.

electricjones:
This looks great. Merging.

PR #111, opened 2016-01-02
Simplifies builder tests dramatically

electricjones:
This is a PR against the Multiple CRUD. I didn't just push the commits because it has wider implications.

I noticed (while creating scenarios) that the Builder and BaseBuilder tests were doing way too much work (and the wrong work). They shouldn't care about the actual scenarios. All they do is append data to the bag in the predictable way.

So, I rewrote and cleaned up those tests dramatically.

The scenarios belong in the Language Processor Tests and in the QueryBuilder (integration) tests.

Let me know if there are any objections to this. I don't need to merge it right away to keep doing what I'm doing.


smolinari:
No qualms from me! :smile: 

Scott

electricjones:
Always good to have another set of eyes! I've got everything for orient to green, so I'm going to merge this. It can always be undone later as the `develop` branch still has the old Builder tests.

PR #119, opened 2016-04-01
Vagrant ci

electricjones:


electricjones:
Now its failing on vagrant. Ah forget it. Move to docker asap. Leaving develop alone, cause at least it works

PR #121, opened 2016-04-07
Multiple operations

electricjones:
_DO NOT MERGE - This still has a list of todos_

This PR passes all tests using OrientDB. It includes updated unit and integration tests. Right now, Cypher, and Gremlin tests are excluded in `phpunit.xml.dist`.
## What Has Been Done
- [ ] Cleaned up `Bag` and made it ready for mulitple operations
- [ ] Added `$create` and `$update` as arrays to hold data to be created/updated
- [ ] Added `$delete` as a bool to declare whether or not to delete retrieved elements
- [ ] Renamed `$projections` to `$retrieve` for consistency
- [ ] Put the responsibility for validating a `Bag` on the `Bag` itself by adding `prepare()` which is intended to ensure a `Bag` is ready to be processed. (Note there is no validation or defaulting yet, just a list of things to check)
- [ ] Updated `BaseBuilder` to simply populate the new `Bag`
- [ ] Updated `BaseBuilder`'s existing tests, which are all single operation
- [ ] Added `BaseBuilder/ScenariosTest` to test for the five scenarios that kicked off #81 using multiple operations.
- [ ] Revamped the `OrientSqlProcessor` and added an `AbstractProcessor`. This can be added to, if helpful for the other processors

And probably more, lol.
## ToDo BEFORE merge
- [ ] Review Pull Request in General
- [ ] Update Cypher Processor to process new `CommandBag`
  - [ ] UnExclude Neo tests from phpunit.xml.dist
- [ ] Update Gremlin Processor to process new `CommandBag`
  - [ ] UnExclude Gremlin tests from phpunit.xml.dist
  - [ ] Implement Gremlin Query Builder Test (should be super simple)
- [ ] Update processors to handle new Bags _and_ to process injected instances of `BaseBuilder`
- [ ] Clean up tests
  - [ ] Add more scenarios, if needed
  - [ ] Update Builder and maybe Query tests
- [ ] Write documentation
### And some Would be nice
- [ ] use modern_graph in every test and expected result
- [ ] internally document scenarios
## Scenarios and Testing

(updated 4-8-16)
This PR also introduces Scenarios into testing, and generally refactors the tests a bit. The idea is that a new `Scenario` will extend `Spider\Test\Scenarios\AbstractScenario`. So, a scenario may be "insert two vertices and and edge between." Then, throughout the system, the scenario would provide the standard piece for that scenario. Right now, that is just a description and the `CommandBag`, but that could be updated to include expected results, or any number of things. The key (I think) is to keep these scenarios _generic_, so that each test that implements the scenario for a specific driver can provide what it needs.

Scenarios are already implemented in `Spider\Test\Unit\Commands\Languages\BaseTestSuite`. The OrientSQL implementation handles scenarios. The other will have to be brought up to snuff, though that should automatically happen with the new language processors.
### Questions / todos around scenarios and testing
- [ ] Should/Could we implement these same scenarios into the integration testing for `Query` builder?
- [ ] Move `modern_graph` to a central place and use the same expectations throughout
- [ ] Look at simplifying the `Driver` tests. All we really need those to test is connections, transactions, responses, and executing commands. Probably don't need so many "create, write, etc." tests


electricjones:
And I will detail the scenario testing and some other documentation in that first comment. Ran out of time tonight.

dmill-bz:
+4.5k lines. It's like reading a book! ;)

electricjones:
Yeah, she's not a small beast. But, I like the direction.

electricjones:
I just updated the description to include info about scenarios and testing

electricjones:
Now that there is a roughly working docker setup (with some love needed, see #128), I plan on tackling the Cypher Processor next (hopefully this week).

Issue #123, opened 2016-05-11
[App] Introduce a (basic) spider-cli application


electricjones:
At this point, this would be an incredibly simple cli tool, but the idea is that it could be easily extended as time goes by. Using Symfony's Console Component, it would be easy to maintain and build.

At this point, I think all the app would need to do is:
- [x] Manage testing commands (run integration, unit, acceptance tests, configure fixtures, etc)
- [x] Manage `docker` and `docker-compose` commands for fine tuning environment
  (going along with #120)

Some future enhancements may include:
- [ ] Generators for models and the like
- [ ] Inspection utilities so you can connect to your db through spider and mess with data
- [ ] A "tinker" functionality
- [ ] Alert you of updates.
- [ ] Provide some visualization tools for graph data (could go hand in hand with some kind of web-based "studio", though each db includes its own specific tool. May be overkill)

Just some thoughts. Would appreciate feedback.

In any case, I'm going to move forward with the first two in order to simplify development

electricjones:
Going off of PR #125, I propose two different, but related utilities: `spider-dev` and `spider.php`.

`spider-dev.sh` is a bash script that simply manages automated testing and docker setup. It does nothing else.

`php spider.php` extends `spider-dev` and adds the generators, and other functionality.

## Proposed `spider-dev` Api

Across all commands, you may specify which versions you want to run. By default, the most currently supported versions will run.

`--orient:2.1 --neo:x.x --gremlin:x.x, --php:5.6`

`spider-dev test` runs all tests against most recent versions by default
- `-i` Runs integration tests
- `-a` Runs acceptance tests
- `-u` Runs unit tests
- `--phpunit <args>` passes the <args> to phpunit

`spider-dev up` gets all the databases up and running and builds (but does not run) the php container
- `spider-dev up --orient:2.2 --neo:1.7 --gremlin:2.8 --php:7.0`

`spider-dev tinker` runs `up` and then opens the php shell so you can interact with the tests or the code directly

## Proposed API for `php spider.php`

Same as the above (actually just calls `spider-dev`), only included for convenience of using a single application. `php spider.php dev test --orient:1.7`

WILL EVENTUALLY ADD:
- `spider.php generate:model`
- etc

electricjones:
Spider-Dev utility has been merged to develop

Issue #120, opened 2016-04-01
Move from vagrant to docker


electricjones:
Building on a Gitter chat, it is time to move away from vagrant to docker for performance, maintainability, and compatibility with travis.

Especially now that https://blog.docker.com/2016/03/docker-for-mac-windows-beta/ is on the horizon.

Next steps?

dmill-bz:
### 1. Find containers we can use
- gremlin-server (I maintain one that can be used, though I haven't documented it yet)
- neo4j
- orient
- php

All of these should be easy to find, the question of how easy they are to automate is another matter, some selection will be required.

### 2. Build out spider container(s)

Use docker-compose to build and publish our app container. This is a container that will have all the OS + dependencies required to run spider.
This container should allow us to share configuration files with the databases, as well as share the working dir (spider source)

The tricky part here is going to be mix and matching the containers from 1. What I mean by this is that one container can only have one combination of gremlin-server, neo, orient, and php. So if we want to test older versions etc.. Our options will be a little more limited than currently.

### 3. Create scripts in spider for use

This script will basically install and run the docker container from 2. adding source files to the container and providing a bunch of command line arguments for various tasks. We could imagine:
`./build/execute.(sh/bash)` with :
- **`-t`** **`--test`** : runs the full test suit 
- **`-u`** **`--unit`** : runs unit tests only
- **`-i`** **`--integration`** : runs integration tests only
- etc...

smolinari:
Puh, that all sounds awfully complex. That could be and probably is my lack of knowledge though....

Scott

electricjones:
My team uses docker and it's not as bad as it seems.
- [x] Official Orient Container: https://github.com/orientechnologies/orientdb-docker
- [x] Official Neo Container: https://hub.docker.com/_/neo4j/
- [x] Official PHP Container: https://hub.docker.com/_/php/

I would think that the best way to test against different versions is to maintain a couple docker images. We could automate the building of these. Since the dbs themselves don't ever interact, and we've decided only to support certain version, we could essentially have 3, automatically built containers:
- Current versions
- Current versions - 1
- Current versions - 2

And then the cli would have the argument 
- `spider a -t` for the current full test suite
- `spider b -u` for the c - 1 unit tests
- `spider c -i` for the c-2 integration tests

[...]

PR #125, opened 2016-05-14
Migrate to Docker from Vagrant

electricjones:
Picking up on Issues #120 and #123 . **This is not ready to merge yet**

Alright, after a lot of learning, I managed to create a setup that works for orientdb for now. It should be easy enough to add neo and gremlin.

clone the `docker` branch
`docker-compose up -d`
`docker run -it --link spider_orientdb:orientdb --net spider_default -v /Mac/users/michael/code/spider/spider:/spider spider/php bash`

The second command is specific to my system. replace the `/mac/users...` path with the absolute path of spider. (for windows /c/path/to/spider).

This will all be abstracted away beneath a `spider-dev.sh` bash file. After a long chat with a coworker who manages our docker system, I think it'll be best to skip `docker-compose` for now. If we do everything through `docker build` and `docker run` abstracted behind a `spider-dev` bash file, then we give the users the ability to cherry pick and mix-and-match the versions of php and database.

Also, I had to use a php image from `ubuntu:15.10` which I hate. It's huge. I tried several others, but got frustrated b/c I either couldn't install xdebug or they weren't compiled with sockets enabled (I can't figure out why). I am sure there exists, or we can define a better one. For now, though, I just wanted to get it working.
## Still ToDo:
- [x] Remove vagrant
- [x] Setup PHP and Orient to work with shell
- [x] Move fixture and testing credentials to `.env` file (that's why tests fail right now)
- [X] Add Neo4j
- [x] Add Gremlin
- [x] Add a php 5.6 container
- [x] Add versioning for supported versions to `docker-compose.yml` (https://docs.docker.com/compose/compose-file/#variable-substitution)
- [x] Initialize `spider-dev.sh` to abstract away docker commands (from #123)
- [x] Add testing commands to `spider-dev` (`spider-dev test -iau`)
- [x] Integrate with Travis
- [x] Documentation and lots of testing


electricjones:
Neo4j and OrientDB are up and running. I ran into some problems adding the Gremlin container. Will try later this week.

To get this running at this point:

``` bash
docker-compose up
docker run -it --link spider_orientdb:orientdb --link spider_neo4j:neo4j --net spider_default -e "SPIDER_DOCKER=true" -v /Mac/users/michael/code/spider/spider:/spider spider/php bash
```

Replacing the volume path with your own. The next step is to get `docker-compose.yml` to handle these complexities and multiple versions of databases and php.

smolinari:
The only way I can see a `docker-compose.yml` handle different versions is to have different `docker-compose.yml` files with a different base `dockerfile`.  

I've been working with Docker now too the last couple of days and learning a lot. :smile: I've also gotten a PHP image created with Alpine Linux with PHP7 and it has composer + phpunit globally pre-installed. You can get the image with 

`FROM skooppaos/php7`

I can easily create a php5.6 image too. Would we need any other PHP versions?

Scott

electricjones:
That's great! does that php container come with sockets enabled? The last time I tried an alpine container, I couldn't get sockets to work, which is ridiculous. I think 5.6 and 7.0 are all we are really supporting right now. It would be great to have them both installed on the same container, but two different boxes would be fine as well.

And `docker-compose.yml` supports [Variable Substitution](https://docs.docker.com/compose/compose-file/#variable-substitution) which I think should do the trick. Actually, if we push a working php container to hub.docker.io, then we won't need any dockerfiles whatsoever.

smolinari:
I have sockets enabled in the PHP compilation. Is that what you mean?

Scott

electricjones:
Yes, that is exactly what I mean. Thank you.

electricjones:
And, I take it back. Two different tags (`FROM skooppaos/php:7.0`, `FROM skooppaos/php:5.6`) or something similar would be perfect. Also, would you consider making that repo public on github? I'd like to see the Dockerfile, and others may as well.

smolinari:
Ok. I'll push them with those tags. 

It isn't as easy as I thought to make a 5.6 image. Working on it though.

Scott

electricjones:
Your container worked great.

I have also simplified what is needed to run the containers:

`docker-compose run php sleep 15;phpunit` will run all the phpunit tests

The first time you run it, it will pull in and build everything. The "sleep 15" needs a better solution. The php container cannot start running its tests until the db services are up and ready. The best alternative (though more complex than I like) is to have the php bootstrap file check for connections, and hold up the tests until the connections are good.

I'd love to find a better solution.

To open a full terminal, you must still

``` bash
docker-compose up -d
docker run -it --link spider_orientdb:orientdb --link spider_neo4j:neo4j --net spider_default -e "SPIDER_DOCKER=true" -v /Mac/users/michael/code/spider/spider:/spider spider/php bash
```

Also open to a better solution here.

One last note: I have moved all the credentials to environment variables. This means you can use the docker's php container, or use php locally. Simply add a `.env` file to the spider root with the services credentials and you can run the tests and such from your host machine. For me, this was a huge performance boost.

electricjones:
I have pushed a basic shell script to handle docker testing and management. I'm happy with it, though I'm sure it could use some refactoring and testing.

See `bin/spider-dev` for api.

From spider directory `bin/spider-dev test` to run all tests, `bin/spider-dev up` to start services, etc.

**Two important notes**
- If you (like me) are using the docker private beta for mac, you must include the flag `--mac` for now. This will be removed once beta goes out.
- The options MUST come before the command. At least for now.

@smolinari Have you gotten any luck with a php5.6 container? I also noticed that this one doesn't support bash or shell. At least not that I've found. We need a way to do something like

`docker run -it skooppaos/php7 bash` and open up a terminal.

Next step is to finish versioning in the `docker-compose` file, and then add gremlin support.

smolinari:
@chrismichaels84 - I haven't finished the 5.6 version yet. Working on it today. The php7 version is also finished and coincidentally, I also added bash, because of the timeout issue I'm having with ODB (which you also mentioned), I was trying to incorporate [wait-for-it](https://github.com/vishnubob/wait-for-it). So the new image should work for you now. It is also under `skooppaos/php:7.0` as you wished.

I like the script you made. I guess I'll have to test it. :smile:  

Edit: I've also uploaded the files to a repo: https://github.com/skooppaos/docker-files

Edit2: I actually won't be getting the 5.6 version done today. I had too much work on my house and now I have to go shopping for the weekend. So, I'll get to it tomorrow.

Scott

electricjones:
The container looks great. Good work. And wait-for-it could come in handy. I've run into a problem where I have to make containers sleep before running tests b/c we have to wait for the databases to be ready. Something like this could be super helpful. And have a good weekend.

electricjones:
Just pushed update with full-ish versioning control. If you try to use an unsupported version, you will get an error, but it will let you try. If there is a docker image with that tag, it will load.

Also added `spider-dev versions` which prints some info.

Does NOT include actual supported versions. That will have to be its own ticket to see what spider actually supports. This way, though, we get to keep each branch and its supported versions together, which is nice.

Last cleanup items before this gets tested and merged (other than documentation)
- [x] Some simple refactoring
- [x] Better handling of working directories
- [x] Add gremlin container

Would be nice:
- [x] ~Integrate `wait-for-it` so that `spider-dev` doesn't run tests or open a shell until the environment is ready. I first thought to do this is phpunit, but I think those tests _should_ fail if the service is unavailable.~ This deserves its own ticket and is unimportant atm.

My last thought is along with #123, add a `spider.php` cli utility with a command to build this bash file automatically from a `supported_versions.json` file or something similar. We can use the same meta file in Travis or anything else. One place to change whatever.

smolinari:
Just an FYI. Spider can't support ODB 2.2 yet, because the OrientPHP driver doesn't work properly with 2.2, unfortunately. 

https://github.com/Ostico/PhpOrient/issues/71
https://github.com/orientechnologies/orientdb/issues/6143

Scott

smolinari:
The skooppaos/php:5.6 image is finished and pushed to Docker-hub. It took way too long to do too. I had a compile error and had finally found out the 5.6.21 src on Github is broken (bson version wasn't being accepted), so I used the php.net archive. Don't ask me how I ended up with finding that as the answer, but I did. It just took an incredibly long time to get there. 

I'll test out the whole Docker setup tomorrow morning. 

Scott

electricjones:
Great work.

I think all that leaves is the broken gremlin container and getting the integration with travis right.

@PommeVerte could you take a look at the gremlin container? The docker-compose.yml file has two containers commented out. The first one is yours. When it is active, it builds fine, but when we try to run the tests, the gremlin tests fail, saying the connection was refused. The hostname is "gremlin" and pinging "gremlin" and "gremlin:8182" both work fine. Not sure if the server is failing to start for some reason?

For the moment, the gremlin tests are marked as skipped.

electricjones:
Integration with travis is complete, minus the Gremlin container.

HHVM seems to be failing, but its failing on the very first (Travis-specific) command, so it may be a but in their system. I'll look at it tomorrow.

smolinari:
Bad news. Docker sucks under Windows 7. There is no interactive mode, so the php container just stops and nothing happens. I also can't get volumes to work.  I've reworked (back) in a dockerfile, so I can work with the spider_php image, but that also doesn't help. When I run the container with `run -it`, the  `/spider` directory is empty. I've looked up stuff about volumes with Windows and setting them up
with Virtualbox, but nothing seems to work.  I can copy the files into the container, but then phpunit doesn't work. It doesn't show any errors, nothing. I am lost. It seems my only solution to fix this is moving to Win10. Yuck! :cry: 

Edit: Ha! Look what I found. 

http://stackoverflow.com/questions/37194404/phpunit-not-displaying-error-output-in-docker-container

Same issue I have now, when running the container interactively.  

Scott

smolinari:
Ok. From your SE post, I got an idea and now I have some output.

`No Database Credentials Supplied. Please use the supplied docker container or add a .env file.`

Could it be a compatibility issue with phpunit? The global phpunit installed in the PHP containers is 5.0. The one used in spider is 4.8.

Scott

electricjones:
Yep, that's actually my question, lol. I thought I had fixed that, at least on mine.

The credentials exceptions is actually part of spider. I forgot to document that. When you want to use php on your machine and just connect to the databases, you need to add a `.env` file to the spider directory that looks like `tests/.env.docker` but with your credentials (usually "localhost" for the hostnames, everything else should be the same).

If you are getting that error from one of the spider-dev commands (like `bin/spider-dev shell`) then an environment variable is not being set. The cli utility sets `SPIDER_DOCKER` so the bootstrap file in phpunit knows which `env` file to load the db credentials from.

electricjones:
And I figured docker in windows and mac would suck. I was hoping not so much. I'm really waiting for https://beta.docker.com/ to go public. That's full native docker in windows and mac so it works exactly like linux.

smolinari:
Yeah, but the "native docker" in Windows will only work with Win10. :cry: 

Scott

electricjones:
Bummer. I hadn't realized that. The containers should still work, and you should be able to use the databases. But PHP would have to be installed on your computer, or you should be able to setup something like PHPStorm to use the php container to send remote commands.

In either case, even with no interactive capabilities, the `spider-dev test` command should still work.

smolinari:
That's the problem. When I use `spider-dev test`, docker returns with

`ERROR: Interactive mode is not yet supported on Windows.`

I think I have a way to get around this. But, it would entail going back to a dockerfile to build the PHP service.

Scott

electricjones:
That's weird. There is no interactive flag in that command. It's basically the same as running
`docker-compose up`
`sleep 30`
`docker-compose run php vendor/bin/phpunit`

Whenever you get a chance, I'd like to see what you arrive at as a workaround. If needed, we can create a separate `bat` file for windows, update the `sh` file, or just keep vagrant around for windows < 10. I didn't upgrade to 10 either.

electricjones:
This probably doesn't help the windows problem, but a general improvement.

The `test` command now returns the exit code from phpunit, so you can run it in a more automated way.

The travis build fails b/c hhvm does not seem to be supported on docker builds of php. I have to do some research to figure that out. One option is to use `spider-dev test` command AS the test command, now that it returns the correct exit code. Question is, how to manage different php versions for different builds.

electricjones:
I've opened up a ticket on Travis to see if they will support hhvm. The other option (which we should probably do anyway) is to officially add an hhvm image to the list and allow people to choose hhvm from `spider-dev`. Then, we could forgo using travis's language feature altogether and run the tests directly from our own docker container.

Ticket is travis-ci/travis-ci#6091.

electricjones:
Well, Travis CI says hhvm is not supported with docker and they don't plan to. Looks like we should add an hhvm container.

electricjones:
I have added hhvm to the environments, and updated the travis.yml file to use docker for everything. However, for some reason, my committs aren't pushing to github. Super weird.

electricjones:
And that's basically it! I have added HHVM and done some final cleanup. Travis is now integrated to use the docker machines and we can have as many combinations as we like.

All that's left is to get the gremlin container working. I may take a stab at it tomorrow. I don't know why I've been having problems with it :/

@smolinari If we can't get this to work on windows/mac without the native app, then I vote we direct users to your spider-box for those instances. It has everything that's need and is known to work well. I've added some documentation in docs/contributing.md. Feel free to add/edit anything in the spider box section.

smolinari:
Yup, I can't get Docker to do as I want on my Win7 system and I've tried all sorts of things. I am giving up, which I hate to do. I guess I'll stick with Vagrant for now as you suggest. 

The only thing missing is PHP7 support in the spider-box. I'll see if I can add that in the near future.

Scott

electricjones:
Well, this is actually working -- and all finished with gremlin and all -- but there are two external problems:
- [ ] The Gremlin container does not work with transactions. I have opened a ticket PommeVerte/docker-gremlin-server#2. For now, that one test is skipped for gremlin
- [ ] Travis seems to be having trouble with the neo container. However it all works locally.

electricjones:
So, I've run into a really weird problem. It seems that (only on travis) the neo4j tests can't connect to neo4j when the gremlin database is running. Works fine on my local machine. When the gremlin db is disabled, all works well. Even when the gremlin tests are skipped -- if the db is up, then neo4j isn't.

But only on travis.

@PommeVerte any idea what may be causing this?

electricjones:
I have decided to revert to a point where everything is working except gremlin, merge this into develop and then into the other two PRs. This will allow me to use it in real situations for testing.

We just have to be sure not to release develop until the gremlin container is working as expected. If I find myself with time and @PommeVerte isn't able to look at it, I may fork the gremlin container and see what I can do. I need to learn tinkerpop better anyway.

